PR #33, opened 2026-04-04
chore: add comprehensive .gitignore for Rust, IDEs, and OS files

olegshulyakov:


Kuberwastaken:
already shipped, thank you though :)

olegshulyakov:
This one is much more wide and include IDEs and different OS temp files.

PR #49, opened 2026-04-06
fix: restore OpenAI Codex subscription flow

cgasgarth:
## Summary
- restore the ChatGPT Plus/Pro `openai-codex` path end-to-end, including OAuth token refresh, `ChatGPT-Account-Id`, model selection, and runtime provider switching after `/connect`
- align the Codex request/response handling with the official ChatGPT backend shape, including required `instructions`, `store: false`, `stream: true`, and SSE response parsing
- keep the runtime behavior scoped to the Codex provider path so non-Codex providers continue using the existing request flow

## Verification
- `cargo check`
- `cargo build`
- `cargo test codex`
- `cargo test test_anthropic_to_openai_request_basic`
- `cargo test stream_accumulator_text`

## Notes
- I verified the runtime changes are gated to `openai-codex` / `openai-codex/*` selection in the query path.
- A broader existing repo test failure outside this Codex work was observed earlier in `crates/tui/tests/render_snapshots.rs` (`thinking_block_collapsed`), so the verification above is intentionally focused on the Codex path.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+repos:
+  - repo: local
+    hooks:
+      - id: cargo-fmt-check
+        name: cargo fmt --check
+        language: system
+        stages: [pre-commit]
+        pass_filenames: false
+        entry: bash -lc 'cd src-rust && cargo fmt --check'
+      - id: cargo-check
+        name: cargo check --locked
+        language: system
+        stages: [pre-commit]
+        pass_filenames: false
+        entry: bash -lc 'cd src-rust && cargo check --locked'
+      - id: cargo-test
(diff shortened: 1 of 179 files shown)


aliensouls:
if one has codex already, why would he use this harness instead of codex itself, what are the features make it worth it? just curious

iongion:
> if one has codex already, why would he use this harness instead of codex itself, what are the features make it worth it? just curious

If one has claude already, why would he use this ?

For fun dude, for learning, for building and creating stuff, not all has economical or logical reasoning, just because is good enough!

jhult:
FYI, this is hard to review because the line endings changed.

You can use these apps to convert: https://dos2unix.sourceforge.io/

PR #67, opened 2026-04-08
fix: make nix dependency Unix-only for Windows compilation

Nullislander0:
## Summary
- Moves the `nix` crate dependency in `crates/cli/Cargo.toml` from unconditional `[dependencies]` to `[target.'cfg(unix)'.dependencies]`
- The code using `nix` in `main.rs:470-475` is already gated behind `#[cfg(unix)]`, but the dep itself was unconditional, which prevents compilation on Windows since `nix` is a Unix-only crate
- Also adds SearXNG as a primary web search backend (3-tier fallback: SearXNG → Brave → DuckDuckGo)

## Test plan
- [x] Full workspace `cargo build` passes on Windows 10 (previously failed due to `nix` crate)
- [x] Unix `#[cfg(unix)]` block for root detection remains unchanged
- [x] `crates/tools` already had `nix` correctly gated — only `crates/cli` needed the fix
- [ ] Verify Linux/macOS CI still passes (no functional change on Unix)

### DIFF
--- a/src-rust/crates/cli/Cargo.toml
+++ b/src-rust/crates/cli/Cargo.toml
@@ -34,11 +34,13 @@ url = { workspace = true }
 crossterm = { workspace = true }
 parking_lot = { workspace = true }
 dirs = { workspace = true }
-nix = { workspace = true }
 base64 = "0.22"
 sha2 = "0.10"
 open = "5"
 urlencoding = "2"
+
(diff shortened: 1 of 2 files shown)


Nullislander0:
Splitting into two separate PRs — one for the Windows fix, one for SearXNG.

PR #69, opened 2026-04-08
feat: add SearXNG as primary web search backend

Nullislander0:
## Summary
- Adds SearXNG (self-hosted meta-search engine) as the preferred web search backend
- Three-tier fallback: SearXNG → Brave Search API → DuckDuckGo Instant Answers
- SearXNG aggregates results from multiple engines (Google, Bing, DDG, Brave, etc.) and returns engine attribution and relevance scores
- Configured via `SEARXNG_URL` env var (defaults to `http://localhost:8888`)
- Bumps max results from 10 to 20
- Gracefully falls back to existing Brave/DDG path if SearXNG is unavailable (15s timeout)

This keeps the existing search behavior fully intact for users without SearXNG — the only change for them is the higher max results cap.

## Test plan
- [x] `cargo build -p claurst-tools` compiles clean
- [x] Full workspace build passes on Windows 10
- [ ] Test with SearXNG running locally (`SEARXNG_URL=http://localhost:8888`)
- [ ] Test fallback when SearXNG is down (should fall through to Brave/DDG as before)
- [ ] Test with `BRAVE_SEARCH_API_KEY` set (second tier)
- [ ] Test with no env vars set (DDG fallback, existing behavior)

### DIFF
--- a/src-rust/crates/tools/src/web_search.rs
+++ b/src-rust/crates/tools/src/web_search.rs
@@ -1,15 +1,18 @@
-// WebSearch tool: search the web using Brave Search API or fallback to DuckDuckGo.
+// WebSearch tool: search the web using SearXNG (preferred), Brave Search API,
+// or DuckDuckGo Instant Answers as a fallback.
 //
-// Mirrors the TypeScript WebSearch tool behaviour:
-// - Accepts a query string
-// - Returns a list of results with title, url, and snippet
-// - Falls back to DuckDuckGo if no search API key is configured
+// Priority order:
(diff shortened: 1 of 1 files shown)


Kuberwastaken:
Closing as superseded: main has had a SearXNG backend since 59a3300 (`SEARXNG_URL` -> Brave -> DuckDuckGo, see docs/tools.md "WebSearch"). It differs from this PR in being opt-in — it never probes localhost:8888 unless `SEARXNG_URL` is set — which is the behaviour I want.

The one idea from here worth adding is falling through to Brave/DDG when a configured SearXNG instance errors. If you want to send that as a small PR against current `web_search.rs` with a test, happy to take it. Thanks for the original push on this.

PR #70, opened 2026-04-08
feat: add configurable custom OpenAI-compatible provider

k99k5:
add configurable custom OpenAI-compatible provider

### DIFF
--- a/src-rust/crates/api/src/providers/openai_compat_providers.rs
+++ b/src-rust/crates/api/src/providers/openai_compat_providers.rs
@@ -6,6 +6,7 @@
 // variable is absent or empty the provider is still constructed but
 // `health_check()` will return `ProviderStatus::Unavailable`.
 
+use claurst_core::Settings;
 use claurst_core::provider_id::ProviderId;
 
 use super::openai_compat::{OpenAiCompatProvider, ProviderQuirks};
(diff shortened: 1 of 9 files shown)


k99k5:
@Kuberwastaken

JackCloudman:
Hey, nice feature! I took a look at the implementation and spotted a few things worth addressing before merge:

**Critical**

**1. Double-write in `persist_custom_provider_base_url` breaks settings merge logic** (`app.rs:1482`)
The function writes to both `settings.providers` *and* `settings.config.provider_configs` via `insert`. This unconditionally overwrites `provider_configs`, which destroys the precedence that `effective_config()` establishes with `or_insert_with`. The fix is to drop the `settings.config.provider_configs.insert(...)` line entirely — `settings.providers` is the correct single source of truth here.

**2. `can_submit` blocks local endpoints that don't need an API key** (`custom_provider_dialog.rs:87`)
The default URL is `http://localhost:11434/v1` (Ollama), but submit is blocked until an API key is entered. The three local providers already in the project (Ollama, LM Studio, llama.cpp) all work without one. The API key field should be optional — only the URL should be required for submit.

**Minor**

- `custom_openai()` calls `Settings::load_sync()` at construction time, introducing hidden disk I/O in the registry hot path. All other providers in that file only read env vars. Worth splitting into `custom_openai_with_url(base_url)` (no I/O) for the registry path.
- `move_prev_field` is an alias of `move_next_field` — works fine with 2 fields, but will silently break if a third field is ever added.
- Indentation on the `masked_key` block won't pass `rustfmt`.
- The "Popular" category for this item in the provider picker feels off — "Custom" or "Advanced" would be more accurate.

Overall the dialog structure is clean and follows existing patterns well. The API key masking and the `take_values()`/`close()` combo are done correctly. Just the two critical items above would be worth fixing before this lands.

Kuberwastaken:
@JackCloudman great breakdown, thank you! would be happy to merge if these are included / corrected :)

k99k5:
@Kuberwastaken I fixed it.

AdamBajger:
@k99k5 I believe your issue is already fixed in the main branch, can you report if you still have the same issue when building from main, so we can proceed in some way?

k99k5:
> @k99k5 I believe your issue is already fixed in the main branch, can you report if you still have the same issue when building from main, so we can proceed in some way?

Not fixed. I messed up this PR by mistake. I'll just open a new one.

k99k5:
#97

PR #73, opened 2026-04-08
Fix ChatGPT Codex provider integration

sovereign313:
This fixes several issues in the ChatGPT Codex provider path:

  - normalize provider selection so Codex uses the correct provider id
  - clean up Codex model selection/default model handling
  - send top-level `instructions` from the system prompt
  - send `stream: true` on the streaming path
  - remove unsupported `max_output_tokens` from Codex requests
  - handle the actual Codex SSE event flow, including `response.output_text.delta` and `response.output_text.done`

  Context:
  The adapter was previously shaped more like a generic Responses client, but the live ChatGPT Codex endpoint behaves differently in a few important ways:
  - it requires `instructions`
  - it requires `stream: true` for the streaming path
  - it rejects `max_output_tokens`
  - streamed text arrives via `response.output_text.delta` / `done`, not via `response.completed.output`

  This patch is based on testing against the live endpoint and fixes the blank/no-output behavior that followed from the old parser assumptions.

### DIFF
--- a/src-rust/crates/api/src/providers/codex.rs
+++ b/src-rust/crates/api/src/providers/codex.rs
@@ -24,7 +24,7 @@ use claurst_core::codex_oauth::{
 use claurst_core::oauth_config::{get_codex_tokens, save_codex_tokens, CodexTokens};
 use claurst_core::provider_id::{ModelId, ProviderId};
 use claurst_core::types::UsageInfo;
-use futures::Stream;
+use futures::{Stream, StreamExt};
 use serde_json::{json, Value};
 use tracing::{debug, warn};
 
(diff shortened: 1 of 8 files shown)


Kuberwastaken:
@sovereign313 have you tested locally? looks decent but I've ran out of codex credits so these are hard for me to verify lol

sovereign313:
I have tested it, since I use codex way more than claude. 

I was about to close it, since I saw another further down (3 days ago) that aims to fix it too, but I haven't looked at theirs.

Kuberwastaken:
@sovereign313 would be awesome if you could compare the two/ work on a single one - it's hard for me to test codex hence the wait haha

sovereign313:
They're pretty close to the same, but mine has a lot more things in it. The other is more focused to just fixing codex with SSE and stream.  Mine does that plus provider state and sync and added PRIMARY paste.  Honestly, It would be cleaner to merge the other one, and close mine, and have me resubmit just the PRIMARY paste addition.  Mine does too much at once probably.

Kuberwastaken:
@sovereign313 primary paste was something I was gonna intend to work on anyway, honestly lgtm

Issue #48, opened 2026-04-06
Custom base URL and API key not loading/injecting properly


AdamBajger:
I am using a custom provider with a custom base URL for requests and a custom API key issued by them.
The provider is OpenAI API compatible, but Claurst does not really offer a way to set custom base url.

Furthermore, even after I got the custom base URL working by tweaking the hardcoded fallbacks and default api endpoints (ugh, horrible!)
the system could not spawn subagents using that base URL and only whined about ANTHROPIC_API_KEY not being set.

I believe there should be a clean system of preferring the specified base url and api key if provided.

May I try to fix this, despite knowing little to zero Rust? :D

Kuberwastaken:
Should be a simple fix! Yes, feel free to try to fix it :)
lmk if you want me to do it instead @AdamBajger

AdamBajger:
Hmm, interesting. The issue did not seem to be resolved when I tried to use the latest version 2 days ago. I Have created my own PR to fix this properly.

AdamBajger:
Is this not resolved?

Kuberwastaken:
should be fixed, idk why I reopened it lol

PR #74, opened 2026-04-08
fix: api_base

whilb:
Had trouble configuring claurst for my gb10. 

summary:
- Fix double /v1 in custom base URLs
- Fix api_base override getting silently ignored — The override was applied to registry_provider, then immediately discarded if a runtime provider existed. Moved the override to happen after that resolution so it actually does something.
- Add llama-server as a provider alias to match registry
- Dropped the /v1 from the Ollama example URL since the code adds it anyway.

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -257,7 +257,7 @@ Connects to a locally running Ollama instance. No API key required.
   "provider": "ollama",
   "providers": {
     "ollama": {
-      "api_base": "http://localhost:11434/v1"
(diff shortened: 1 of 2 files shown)


whilb:
Likely related to https://github.com/Kuberwastaken/claurst/issues/48

Kuberwastaken:
Yeah intended to fix but @AdamBajger wanted to help with a first PR haha, lgtm tho

PR #83, opened 2026-04-10
docs: add fish shell PATH instructions and expand Apple Silicon support to M4/M5/A-series

maxffarrell:
## Summary

- **Fish shell**: adds a dedicated *Adding to PATH* section for fish using `fish_add_path`, explains universal variable persistence (no restart needed), and includes a `config.fish` alternative
- **Fish completions**: adds a fish snippet to the Shell Completions section
- **Apple Silicon**: updates the system requirements table and install section heading to cover M4, M5, and A-series chips (e.g. MacBook Neo); clarifies that all Apple Silicon Macs use the same `aarch64` binary
- **Bug fix**: the macOS install one-liners referenced `claurst` after extraction, but the archive actually extracts to `claurst-macos-{arch}` — commands now use the correct filename and include a final `mv` so the binary lands on `PATH` in one paste

## Test plan

- [ ] Verify fish `fish_add_path` instructions work on a fresh fish install
- [ ] Confirm `claurst-macos-aarch64` binary name matches what the release archive extracts
- [ ] Check that M4/M5 Mac users can follow the aarch64 install path without confusion

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -14,10 +14,14 @@ from source.
 | Linux    | x86_64      | glibc 2.17+ (most distros from 2014 onward) |
 | Linux    | aarch64     | glibc 2.17+ (Raspberry Pi 4, AWS Graviton, etc.) |
 | macOS    | x86_64      | macOS 11 Big Sur |
-| macOS    | aarch64     | macOS 11 Big Sur (Apple Silicon: M1/M2/M3) |
+| macOS    | aarch64     | macOS 11 Big Sur (Apple Silicon: M1/M2/M3/M4/M5 and all A-series chips) |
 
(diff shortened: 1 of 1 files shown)


Kuberwastaken:
Closing this draft: `docs/installation.md` has been rewritten around `install.sh` / `install.ps1`, which already handle fish (`.config/fish/config.fish`), and the curl snippets here use `claurst-macos-aarch64` as the extracted filename while the archives contain a plain `claurst`.

If you still want the manual-install section to mention `fish_add_path ~/.local/bin` and to drop the "M1/M2/M3" wording for plain "Apple Silicon", a small PR against current main with just those edits is welcome.

Issue #87, opened 2026-04-11
Subagents refused without Anthropic Key


joebitcoinorg:
version claude 0.0.9

I was asking it to work on a project all night with subagents.  
I'm using the z.ai  support recently merged into main. 

It said an error about not being able to perform the subagent tasks without something like ~"CLAUDE_ANTHROPIC_KEY". 

Although I was able to proceed in a solo developer mode.

Kuberwastaken:
@joebitcoinorg interesting, I've been working with subagents on Gemini and it's working haha

Will look into it!

AdamBajger:
I have ran into this issuse when preparing a fix for custom base_url for OpenAI-compatible providers. I have opened a [PR](https://github.com/Kuberwastaken/claurst/pull/89) to fix this, but I did not have resources to go through the codebase for other providers (I would not be able to access the other providers' APIs)

joebitcoinorg:
> I have ran into this issuse when preparing a fix for custom base_url for OpenAI-compatible providers. I have opened a [PR](https://github.com/Kuberwastaken/claurst/pull/89) to fix this, but I did not have resources to go through the codebase for other providers (I would not be able to access the other providers' APIs)

I could give you an api key for z.ai.  It's not the pro one, but it will give you up to GLM-4.7 .  My $10 subscription expires on the 14th of April. 

oh, and I think the plan might only allow simultaneous api calls on the GLM-4.5 model.   I don't have the pro subscription.

joebitcoinorg:
> I have ran into this issuse when preparing a fix for custom base_url for OpenAI-compatible providers. I have opened a [PR](https://github.com/Kuberwastaken/claurst/pull/89) to fix this, but I did not have resources to go through the codebase for other providers (I would not be able to access the other providers' APIs)

A key is below.  Z.ai  use GLM-4.5 .   The provider might not allow simultaneous api calls... if they do, it will be with the 4.5 model (their latest is GLM-5.1). 

---deleted key provided for testing ----

AdamBajger:
I'll try to get to that. Have you tried latest release in the meantime? Are you sure the issue in not fixed already? 2 hours ago my PR was merged, it might have solved this issue! I will probably have time only later today, in the evening and that is still not 100% sure.

Kuberwastaken:
Should ideally be fixed! Would love for someone to check
Thanks for the PR @AdamBajger

joebitcoinorg:
[...]

PR #89, opened 2026-04-11
feat: enable custom OpenAI compatible providers with custom models

AdamBajger:
- add a dynamic resolution and overrides based on provided base url and api key to the hardcoded OpenAI-based providers
- fix provider delegation in sub-agent tool to inherit the custom openai provider from main agent (tested only for the custom base URL of that one service I have)
- move the openai-compatible provider resolution to a separate file (improved readability)
- allow specifying a custom model string in the model picker, if the hardcoded selection does not match the provider (otherwise, the custom base URL will not be useful if there are different model IDs in the custom provider, and I also find it a no brainer to have this)
- fix missing dependencies required for build

All of these are minimal changes to make the custom provider work.

### DIFF
--- a/src-rust/Cargo.lock
+++ b/src-rust/Cargo.lock
@@ -870,6 +870,7 @@ dependencies = [
  "enigo",
  "futures",
  "glob",
+ "image",
  "nix 0.29.0",
  "once_cell",
  "open",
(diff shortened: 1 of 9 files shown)


AdamBajger:
I believe this fixes #48, and fixes #87 as well.

Kuberwastaken:
Fire

PR #93, opened 2026-04-12
fix:  background agent handling and monitoring instructions

### LINKED ISSUE #87: Subagents refused without Anthropic Key

AdamBajger:
I have went through the issue with subagents and verified that the z.ai API works. I identified a potential culprit causing users to feel like the subagents are not working - monitoring of the bg tasks does not work - the main thread does not see the agents.

This PR should correct mismatches in handling background tasks at least to the extent that the subagents are now monitored correctly.

 I have also check a subset of other tools to make sure other tools did not get broken.
 
 closes #87


Kuberwastaken:
LGTM

PR #100, opened 2026-04-14
feat: add devcontainer setup with Dockerfile and configuration

AdamBajger:
- Add a Dockerfile from base Rust image with a few tools required by the build
- Add a devcontainer.json to enable easy development setup on any machine with DOcker installed
- The setup includes SSH and GPG forwarding, given the user has fullfilled the prerequisites on the host machine
- .claurst is mounted to the project directory to enable quick access to the setting and session history

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,5 @@
+FROM rust:1-bullseye
+
+RUN apt-get update \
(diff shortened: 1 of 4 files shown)


Kuberwastaken:
Fire

Kuberwastaken:
Officially PR #100 🙌

PR #110, opened 2026-04-21
feat: wire PermissionManager into the interactive permission flow

k99k5:
## Summary
  - route TUI and tool permission checks through `PermissionManager`
  - add queued interactive permission handling in the TUI for pending approvals
  - enforce workspace-aware read boundaries and MCP/file-related permission checks in default mode

  ## Test plan
  - [x] cargo build --package claurst
  - [x] cargo test --workspace
  - [x] manually verify interactive permission prompts in TUI

  ## Image
PixPin_2026-04-21_11-34-44

### DIFF
--- a/src-rust/crates/tui/src/app.rs
+++ b/src-rust/crates/tui/src/app.rs
@@ -5599,6 +5599,7 @@ mod tests {
             "tu-1".to_string(),
             "Bash".to_string(),
             "wants to run".to_string(),
+            "wants to run".to_string(),
             "git status".to_string(),
             Some("git".to_string()),
         );
@@ -5631,6 +5632,7 @@ mod tests {
             "tu-2".to_string(),
             "Bash".to_string(),
             "wants to run".to_string(),
+            "wants to run".to_string(),
             "cargo build".to_string(),
             Some("cargo".to_string()),
         );
@@ -5661,6 +5663,7 @@ mod tests {
         let pr = PermissionRequest::bash(
             "tu-3".to_string(),
             "Bash".to_string(),
+            "wants to run".to_string(),
             "wants to run".to_string(),
             "npm install".to_string(),
             Some("npm".to_string()),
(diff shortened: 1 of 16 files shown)


AdamBajger:
@Kuberwastaken this looks nice. It will probably do my best to integrate this into my work on Kairos mode, as I ran into the issue with permissions myself :D (I let Sonnet design that instead of Opus and the model completely skipped it, lol)
If there are no conflicts, I would merge this and then take it further from there.

src-rust/crates/tui/src/app.rs
```diff
@@ -5599,6 +5599,7 @@ mod tests {
             "tu-1".to_string(),
             "Bash".to_string(),
             "wants to run".to_string(),
+            "wants to run".to_string(),
```
AdamBajger: seems like a duplicate, is it correct?

src-rust/crates/tui/src/app.rs
```diff
@@ -5631,6 +5632,7 @@ mod tests {
             "tu-2".to_string(),
             "Bash".to_string(),
             "wants to run".to_string(),
+            "wants to run".to_string(),
```
AdamBajger: seems like a duplicate, is it correct?

AdamBajger: I see, one is ddescription and the second is reason. Not sure why both is needed tho. We can change it later anyways.

src-rust/crates/tui/src/app.rs
AdamBajger: resolved

AdamBajger (approved):
Yeh, looks good to me, roughly. I suggest merging.

src-rust/crates/tui/src/app.rs
k99k5: I'll clean up the duplicate

Kuberwastaken: would love to merge if the duplicate is confirmed fixed, please ping me :)

k99k5:
@AdamBajger @Kuberwastaken resolved

PR #111, opened 2026-04-24
fix(deepseek): preserve reasoning_content for V4 thinking-mode multi-turn

P1-103n1x:
## Summary

DeepSeek V4 (`deepseek-v4-pro` / `deepseek-v4-flash`) enables thinking mode by default and streams CoT as `reasoning_content`. Per the [official contract](https://api-docs.deepseek.com/guides/thinking_mode), any assistant turn that produced a tool call **must** have its `reasoning_content` echoed back on subsequent turns, otherwise the server returns:

```
400: The `reasoning_content` in the thinking mode must be passed back to the API.
```

The OpenAI-compatible provider path discarded reasoning in three places, making all multi-turn tool flows against V4 fail. This PR fixes all three.

## What changed

1. **`providers/openai_compat.rs`** — stream loop now opens a dedicated Thinking block (reserved non-colliding index) on the first reasoning delta and closes it before text / tool_calls / finish_reason. Previously `ReasoningDelta` was yielded without a matching `ContentBlockStart`, so `StreamAccumulator` had no partial to append into and the reasoning was silently dropped.

2. **`query/src/lib.rs`** — the non-Anthropic provider's inline accumulator only handled Text and ToolUse; `ReasoningDelta` / `ThinkingDelta` fell through to `_ => {}` and were lost. Reasoning is now collected into `reasoning_chunks` and prepended as a `ContentBlock::Thinking` on the assistant message.

3. **`providers/openai.rs`** — `assistant_content_to_openai` dropped `Thinking` on outbound. It now returns the reasoning alongside text / tool_calls, and `to_openai_messages` attaches it as the assistant message's `reasoning_content` field. Messages carrying only reasoning (no content, no tool_calls) fall back to `content: ""` since DeepSeek rejects `content: null && tool_calls: []`.

Unknown fields are ignored by providers that don't recognize `reasoning_content`, so this is safe for OpenAI / Azure / Groq / etc. — no conditional per-provider gating required.

## Test plan

- [x] Unit tests compile (`cargo build --release --package claurst`)
- [x] Manual: `claurst -p "Use Glob on <dir>/*.md then count" --max-turns 4 --permission-mode bypass-permissions` against `deepseek-v4-pro` — multi-turn tool flow completes correctly (previously failed with the 400 above).
- [x] Verified the exact multi-turn contract against the live DeepSeek API: request with `reasoning_content` preserved → 200; without → 400 matching the rule.
- [x] Maintainer to confirm no regression against Anthropic / other providers (tests covering other providers pass locally).

## References

- DeepSeek thinking-mode API contract: https://api-docs.deepseek.com/guides/thinking_mode
- DeepSeek V4-Pro encoding rules: https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro/blob/main/encoding/README.md

Signed-off-by: Phoenix Ye &lt;P1-103n1x@proton.me&gt;

### DIFF
--- a/src-rust/crates/api/src/providers/openai.rs
+++ b/src-rust/crates/api/src/providers/openai.rs
@@ -141,17 +141,33 @@ impl OpenAiProvider {
                     Self::append_user_messages(&mut result, &msg.content);
                 }
                 Role::Assistant => {
-                    let (text_content, tool_calls) =
+                    let (text_content, tool_calls, reasoning_content) =
                         Self::assistant_content_to_openai(&msg.content);
                     let mut obj = serde_json::Map::new();
                     obj.insert("role".into(), json!("assistant"));
-                    if let Some(tc) = text_content {
-                        obj.insert("content".into(), json!(tc));
-                    } else {
-                        obj.insert("content".into(), Value::Null);
+                    // If the assistant turn has no visible content and no tool calls
+                    // but does carry reasoning, some OpenAI-compat servers (DeepSeek)
+                    // still require a non-null `content`.  Emit an empty string in
+                    // that case so the message passes validation.
+                    match (text_content, tool_calls.is_empty(), reasoning_content.is_some()) {
+                        (Some(tc), _, _) => {
+                            obj.insert("content".into(), json!(tc));
+                        }
+                        (None, true, true) => {
+                            obj.insert("content".into(), json!(""));
+                        }
+                        _ => {
+                            obj.insert("content".into(), Value::Null);
+                        }
                     }
                     if !tool_calls.is_empty() {
                         obj.insert("tool_calls".into(), json!(tool_calls));
+                    }
+                    // Preserve reasoning_content for providers that require it in
+                    // multi-turn thinking mode (e.g. DeepSeek V4).  Servers that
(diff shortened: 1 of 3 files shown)


jhult:
@P1-103n1x, could you rebase or resolve the conflicts? Then hopefully @Kuberwastaken could get this merged.

kerta1n:
Would like to see this PR get merged, bump

Kuberwastaken:
@kerta1n as @jhult said, would love to merge if you could resolve conflicts :)

P1-103n1x:
Thanks for the heads up. I don’t have much time to properly rebase and re-verify this right now, and I noticed `main` already has related DeepSeek reasoning/tool-call handling：）

I’ll close this PR for now to avoid blocking things. If I later find that the original `reasoning_content` tool-call loop issue is still reproducible, I’ll open a smaller follow-up PR against current `main`.

P1-103n1x:
Also, Anthropic banned my Max account two days ago, so my ability to test this properly is currently somewhere between “limited” and “ceremonial laptop decoration.” 😅

jhult:
This was intended to fix https://github.com/Kuberwastaken/claurst/issues/121

Issue #115, opened 2026-04-26
Ollama returns ERROR after an answer that needed to use a tool.


Thomot512:
<details><summary>settings.json</summary>
<p>

```json
{
  "config": {
    "api_key": null,
    "model": "ollama/qwen3.6:35b",
    "max_tokens": null,
    "permission_mode": "default",
    "theme": "default",
    "output_style": null,
    "auto_compact": false,
    "compact_threshold": 0.0,
    "verbose": false,
    "output_format": "text",
    "mcp_servers": [],
    "lsp_servers": [],
    "allowed_tools": [],
    "disallowed_tools": [],
    "env": {},
    "enable_all_mcp_servers": false,
    "custom_system_prompt": null,
    "append_system_prompt": null,
    "disable_claude_mds": false,
    "project_dir": null,
    "workspace_paths": [],
    "additional_dirs": [],
    "hooks": {},
    "provider": "ollama",
    "provider_configs": {},
    "formatter": {},
    "commands": {},
    "agents": {},
    "skills": {
      "paths": [],
      "urls": []
    },
    "managed_agents": null
  },
  "version": null,
  "projects": {},
  "remoteControlAtStartup": false,
  "permissionRules": [],
  "enabledPlugins": [],
  "disabledPlugins": [],
  "hasCompletedOnboarding": true,
  "lastSeenVersion": "0.0.9",
  "provider": "ollama",
  "providers": {},
  "commands": {},
  "formatter": {},
  "agents": {},
  "skills": {
    "paths": [],
    "urls": []
  },
  "managed_agents": null
}
```

</p>
</details> 

<details><summary>System</summary>
<p>

```nushell
fastfetch
                  -`                     thomot512@CosmicAMDTower
                 .o+`                    ------------------------
                `ooo/                    OS: Arch Linux x86_64
               `+oooo:                   Kernel: Linux 6.19.14-arch1-1
              `+oooooo:                  Uptime: 6 hours, 16 mins
              -+oooooo+:                 Packages: 5 (flatpak-system), 10 (flatpak-user), 1663 (pacman)
            `/:-:++oooo+:                Shell: nushell 0.112.2
           `/++++/+++++++:               Display (LC49G95T): 5120x1440 in 49", 240 Hz [External]
          `/++++++++++++++:              DE: COSMIC 1.0.0
         `/+++ooooooooooooo/`            WM: cosmic-comp (Wayland)
[...]

PR #118, opened 2026-04-27
fix(ollama): replace null content with empty string to prevent crash after tool use

whiteaxesing:
## Problem

When using Ollama with models that use tool calls (e.g. qwen3.5, devstral), 
Claurst crashes after the first tool use with:

  ERROR Provider stream failed provider=ollama 
  error=[ollama] Invalid request: invalid message content type: <nil>

This happens because when an assistant message contains only tool_calls 
(no text), the content field is set to `null`. The OpenAI spec allows this, 
but Ollama rejects it.

## Fix

In `build_messages()`, replace `content: null` with `content: ""` for 
assistant messages when the provider does not require an API key (Ollama, 
LM Studio, llama.cpp).

## Testing

Reproduced and verified on:
- macOS Apple M5, 32GB RAM
- Ollama 0.13+
- Models: qwen3.5:9b, qwen3.5:27b, devstral-small-2:24b

Before fix: crash after any tool use in the same session
After fix: conversation continues normally after tool use ✅

### DIFF
--- a/src-rust/crates/api/src/providers/openai_compat.rs
+++ b/src-rust/crates/api/src/providers/openai_compat.rs
@@ -251,8 +251,11 @@ impl OpenAiCompatProvider {
                 &request.messages,
                 field,
             );
-            // DeepSeek rejects `content: null` on assistant messages — it
-            // requires either a non-null content (even "") or tool_calls.
+        }
+
+        // Some providers (DeepSeek, Ollama) reject `content: null` on
+        // assistant messages — replace with an empty string.
+        if self.quirks.reasoning_field.is_some() || self.quirks.no_api_key_required {
             Self::ensure_content_not_null(&mut messages);
         }
 


jhult:
@whiteaxesing - can you rebase or resolve conflicts?

jhult:
This would fix https://github.com/Kuberwastaken/claurst/issues/115

whiteaxesing:
@jhult done! ✅ 

Hey! Since there were already some changes to the same area of the code, I wanted to make sure my fix was still necessary. I tested the binary built from main (without my changes) and was able to reproduce the crash consistently 👀 

After a tool-use turn, the next prompt would fail with invalid message content type: <nil>. After applying my fix, the error is gone and follow-up messages work correctly. Attaching screenshots as evidence.

Also took the opportunity to clean up the implementation — instead of a manual loop, the fix now reuses the existing ensure_content_not_null helper that was already there for DeepSeek, so both providers are covered by the same function 😄 🙌🏻 🙌🏻 


Before my changes:
Captura de pantalla 2026-05-05 a la(s) 11 06 11 p  m


After my changes:
Captura de pantalla 2026-05-05 a la(s) 11 21 48 p  m

Tested with 

qwen3.5:9b

jhult:
Awesome, thank you @whiteaxesing.

@Kuberwastaken, can you please merge this?

Kuberwastaken:
Awesome! LGTM thanks @whiteaxesing and @jhult

Issue #104, opened 2026-04-16
Mouse capture breaks native text selection, copy, and causes scroll lag


Plaidmustache:
# Mouse capture breaks native text selection, copy, and causes scroll lag

## Environment
- **Claurst version:** 0.0.9
- **Terminal:** Ghostty 1.3.1 (also affects other terminals that honor mouse reporting)
- **OS:** macOS, Apple Silicon
- **Also affects:** Any terminal emulator that respects application mouse reporting (iTerm2, WezTerm, Kitty, Alacritty, etc.)

## Summary

Claurst unconditionally enables full mouse capture on startup via `crossterm::event::EnableMouseCapture`. This prevents the terminal emulator from handling mouse events natively, which breaks text selection, copy/paste, and introduces scroll lag.

### Text selection is broken
Click-drag to highlight text does not produce a real terminal selection. Instead, Claurst renders its own visual "highlight" that stays pinned to screen coordinates while content scrolls underneath it. The highlight detaches from the text, making it impossible to select content beyond the visible viewport.

### Copy does not work
Since the "highlight" is not a real terminal selection, `⌘C` / Ctrl+C does not copy the highlighted text to the clipboard. Claurst has its own internal copy path via right-click context menu, but this is not intuitive and many users expect native terminal copy behavior.

### Scroll is laggy
Every scroll wheel event goes through: terminal → escape sequence → crossterm parses → Claurst processes → viewport re-render → visible lag. In terminals with native scrollback, scrolling is near-instant because the terminal handles it directly.

### Ghostty `mouse-shift-capture` does not help
Ghostty has a `mouse-shift-capture = true` config option that is supposed to let Shift+drag bypass mouse capture for native selection. This does not work with Claurst, likely because the combination of `1003h` (any-event tracking) + `1006h` (SGR encoding) sent by `EnableMouseCapture` prevents Ghostty from intercepting the events before the app.

## Root cause

In `src-rust/crates/tui/src/lib.rs`, `setup_terminal()` calls:

```rust
execute!(stdout, EnterAlternateScreen, EnableMouseCapture)?;
```

`crossterm::event::EnableMouseCapture` sends **all** mouse tracking modes simultaneously:
- `?1000h` — normal tracking (scroll, basic click)
- `?1002h` — button-event tracking (click + drag)
- `?1003h` — **any-event tracking** (captures mouse movement even without button press)
- `?1006h` — SGR mouse encoding
- `?1015h` — URXVT mouse encoding

[...]

PR #120, opened 2026-04-27
fix: clipboard paste over SSH (Windows Terminal / headless Linux)

watrworld:
## Summary

- **Right-click paste now works over SSH**: When `$SSH_TTY`/`$SSH_CLIENT` is set, `EnableMouseCapture` is skipped so the terminal emulator's right-click context menu remains functional. This is why Claude Code CLI worked on the same terminal but claurst didn't — it wasn't capturing mouse events.
- **No more auto-submit on paste**: Windows Terminal injects clipboard text as raw key events over SSH (not bracketed paste). The trailing `\r\n` from Windows clipboard line endings was firing as `Key(Enter)` and submitting the prompt immediately. Rapid-input burst detection (consecutive keys arriving < 15ms apart) now converts burst Enters to literal newlines so the user can review and edit before sending.
- **Ctrl+V modifier fix**: Changed `modifiers.contains(CONTROL)` to `modifiers == CONTROL` so Ctrl+Shift+V no longer triggers the clipboard-unavailable notification.

### Additional improvements
- OSC 52 write fallback in `try_copy_to_clipboard` — copy out of claurst works over SSH without `xclip`/`wl-clipboard`
- OSC 52 read attempt in Ctrl+V and right-click paste handlers
- `arboard` native-clipboard feature (optional, no `xclip`/`wl-paste`/`xsel` needed)
- `Event::Paste` handler in `main.rs` event loop (was previously silently dropped by `_ => {}`)
- Context-aware clipboard-unavailable notification: SSH sessions get "Right-click to paste via context menu", local sessions get "Install xclip or wl-clipboard"

## Test plan

- [x] Tested on headless Raspberry Pi accessed via Windows Terminal SSH
- [x] Right-click → Paste: text lands in prompt, no auto-submit
- [x] Ctrl+V with empty clipboard: shows correct SSH-specific notification
- [x] Ctrl+Shift+V: shows same notification (Windows Terminal sends it as plain Ctrl+V over SSH — this is a terminal limitation)
- [x] Normal Enter key still submits as expected
- [x] Build works with `--no-default-features --features native-clipboard` (no ALSA required for Pi)

### DIFF
--- a/src-rust/Cargo.lock
+++ b/src-rust/Cargo.lock
@@ -900,6 +900,7 @@ name = "claurst-tui"
 version = "0.0.9"
 dependencies = [
  "anyhow",
+ "arboard",
  "base64",
  "chrono",
  "claurst-api",
@@ -915,6 +916,7 @@ dependencies = [
  "hound",
  "icy_sixel",
  "image",
+ "libc",
  "once_cell",
  "parking_lot",
  "ratatui",
--- a/src-rust/Cargo.lock
+++ b/src-rust/Cargo.lock
@@ -918,6 +918,7 @@ dependencies = [
  "hound",
  "icy_sixel",
  "image",
+ "libc",
  "once_cell",
  "parking_lot",
  "ratatui",
--- a/src-rust/Cargo.toml
+++ b/src-rust/Cargo.toml
@@ -67,6 +67,7 @@ dashmap = "6"
 indexmap = { version = "2", features = ["serde"] }
 bytes = "1"
 base64 = "0.22"
+libc = "0.2"
 sha2 = "0.10"
 hex = "0.4"
 getrandom = "0.2"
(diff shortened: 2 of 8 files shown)


Kuberwastaken:
Thanks @watrworld. Closing as superseded: main now has a `mouseCapture` setting (#104) instead of SSH auto-detection, handles `Event::Paste` in the live loop, and captures raw key-flood pastes from Windows Terminal over SSH without auto-submitting (#316), so the auto-submit and right-click problems this targeted are fixed on main.

The OSC 52 part can't land as written: `osc52_read` does raw libc reads on stdin underneath crossterm's event reader (input race), the `unsafe` blocks lack SAFETY comments, and dropping the keyboard-enhancement flags over SSH regresses key handling for everyone on SSH. A small follow-up PR adding just an OSC 52 *write* fallback to `try_copy_to_clipboard` (no stdin reads, no terminal-mode changes) would be welcome.

PR #122, opened 2026-04-28
fix: MCP stdio startup and environment propagation

k99k5:
Register plugin-provided MCP servers before the MCP manager is constructed so stdio servers are connected during startup.Preserve configured environment variables when launching MCP stdio child processes.
Resolve stdio commands through which' when available, while keeping the original command as a fallback.Add regression tests for MCP server env deserialization and stdio command env propagation.

### DIFF
--- a/src-rust/Cargo.lock
+++ b/src-rust/Cargo.lock
@@ -800,6 +800,7 @@ dependencies = [
  "url",
  "urlencoding",
  "uuid",
+ "which",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Kuberwastaken:
Thanks @k99k5. This has drifted too far to rebase: main.rs's MCP startup block was rewritten (MCP trust partitioning for project servers, bare mode, reconnect path) and core `Config` changed shape, so both files conflict, and hoisting plugin MCP registration as written would now bypass the project-MCP trust gate. The env propagation part was already on main (`cmd.envs(&config.env)` existed at your merge-base). Closing this one.

Two pieces are still worth landing separately if you want to open fresh PRs against current main:
1. `#[serde(default)]` on `config::Config` so a partial `"config": {"mcp_servers": [...]}` (the shape docs/mcp.md shows) parses — small, with the settings test you wrote.
2. Connecting plugin-provided MCP servers at startup, routed through `mcp_trust::partition_mcp_servers` like the other server sources.

The `which` resolution only matters for Windows `.cmd` shims; if that is the case you hit, file an issue with the exact command so it can be gated on Windows.

PR #126, opened 2026-05-03
feat: add OpenCode Go as a provider option in /connect

ilyasakin:
OpenCode Go is a flat-rate subscription endpoint at opencode.ai that exposes an OpenAI-compatible /v1/chat/completions surface, fronting Kimi, DeepSeek, GLM, MiniMax, and Qwen models. Wire it up as a stock OpenAiCompat provider so users can pick it from the /connect picker and paste their OPENCODE_API_KEY.


Kuberwastaken:
Sure, seems like a worthy addition :)

PR #136, opened 2026-05-14
feat(tui): auto-copy selection to clipboard on mouse release

jhult:
When `auto_copy_selection` is enabled (default true), releasing a drag selection automatically copies the extracted text to the system clipboard and shows a brief notification. This removes the need for a manual `Ctrl+C`/`CMD+C` after highlighting text.


Kuberwastaken:
Not sure about this - will wait for more feedback, I'd personally not want it in my TUI sessions but hopefully if there's enough feedback we can have a toggle for it in settings

jhult:
Maybe `auto_copy_selection` should default to `false`.

If you look at another such agent, there is demand for it (see issue 29030).

PR #138, opened 2026-05-14
Enable Cmd+C / Cmd+V on macOS; enable kitty keyboard protocol

jhult:
This fixes `Cmd+C` copy, `Cmd+V` paste, and makes `Shift+Enter` reporting more reliable on terminals that support the protocol.


Kuberwastaken:
LGTM

PR #140, opened 2026-05-14
fix(tui): show bypass-permissions dialog instead of onboarding when using --yolo

jhult:
When `--yolo` (`--dangerously-skip-permissions`) is used, the `bypass-permissions` confirmation dialog should be the only visible dialog at startup. Previously, if the user had no credentials configured, both the `bypass-permissions` and onboarding dialogs would be shown simultaneously. Since onboarding is rendered after (on top of) `bypass-permissions`, it would visually cover the security gate, preventing the user from accepting or rejecting bypass mode.

Now the bypass-permissions dialog takes absolute priority: if in bypass mode, onboarding is suppressed until after the user explicitly accepts or declines the security gate.

### DIFF
--- a/src-rust/crates/cli/src/main.rs
+++ b/src-rust/crates/cli/src/main.rs
@@ -1706,25 +1706,25 @@ async fn run_interactive(
         app.agent_mode = Some(agent_name.clone());
     }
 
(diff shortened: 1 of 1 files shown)


Kuberwastaken:
Can't believe I missed this, thanks

PR #144, opened 2026-05-14
fix(tui): bypass-permissions dialog only shows once per session

### LINKED ISSUE #142: The `bypass-permissions` dialog doesn't have any persistence logic like the onboarding dialog does
1. The onboarding dialog has `persist_onboarding_complete()` which saves `has_completed_onboarding = true` to the settings file, so it only shows once.

2. The `bypass-permissions` dialog just shows/dismisses in memory — there's no code saving state to prevent it from showing again next time.

When https://github.com/Kuberwastaken/claurst/pull/140 moved onboarding into an `else` block, it set up the gate correctly, but the bypass dialog still needs persistence logic to only ask once per directory (or session, depending on your intent).

You'll need to add something similar to what onboarding has — either:
- Save a `has_accepted_bypass_permissions` flag to settings, or
- Only show it once per session (store state in the App struct)
- Show it every time (current behavior)

@Kuberwastaken (and others), which behavior do we want to target — ask once ever per directory, or once per session?

jhult:
Fixes Kuberwastaken/claurst#142

The bypass-permissions dialog now only shows once when claurst starts with `--yolo` (bypass mode), instead of showing every time. Users still see it in each new session, which is the intended security gate behavior. Added a session-level tracking flag to prevent repeated prompts within the same session.

### DIFF
--- a/src-rust/crates/cli/src/main.rs
+++ b/src-rust/crates/cli/src/main.rs
@@ -1708,10 +1708,13 @@ async fn run_interactive(
 
     // Mirror TS BypassPermissionsModeDialog.tsx startup gate
     // Shown as the highest-priority startup dialog (blocks all other UI).
+    // Only show once per session — subsequent sessions in the same directory
+    // will show the dialog again (not persisted across sessions).
     use claurst_core::config::PermissionMode;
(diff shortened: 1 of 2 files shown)


Kuberwastaken:
Awesome!

PR #152, opened 2026-05-15
fix: upgrade crossterm to 0.29, add keyboard enhancement warning

### LINKED ISSUE #150: Crossterm not detecting Windows Terminal properly -> CLI refuses to run in powershell
Found this bug as I am running windows (among other operating systems, ofc).
The issue should be solved just by upgrading crossterm, but also I would like to be able to run best-effort on the legacy connhost.exe.
Can we do best-effort on unsupported terminals instead of just refusing to run?

AdamBajger:
- bump crossterm to get rid of pwsh.exe and cmd.exe hardcoded fails in crossterm 0.28
- add better warning message
- keep the CLI in best-effort mode also in older terms

fixes #150


Kuberwastaken:
Awesome! LGTM

Issue #117, opened 2026-04-27
Need field 'Authorization' in headers when calling minimax


ArthurZhou:
calling any model from Minimax results in 
```
ERROR Provider stream failed provider=minimax error=[minimax] Error: Authentication error: login fail: Please carry the API secret key in the 'Authorization' field of the request header

Error: API error: [minimax] Error: Authentication error: login fail: Please carry the API secret key in the 'Authorization' field of the request header
```

in openclaw, there's an approach to include api key in headers: [here](https://github.com/openclaw/openclaw/issues/27600)

wondering how can i achieve this in claurst

PR #143, opened 2026-05-14
fix(minimax): use correct Authorization header format for API auth

jhult:
Fixes Kuberwastaken/claurst#117

Minimax requires the API key to be set directly in the Authorization header without the Bearer prefix. Changed MinimaxProvider to use a custom HTTP client that sets Authorization header with just the API key instead of using AnthropicClient which prepends "Bearer".

Warning: I have not tested this as I don't have a Minimax subscription. @ArthurZhou, can you pull this PR, run `cargo run -p claurst`, and test?

PR #154, opened 2026-05-16
fix(minimax): now minimax use 'X-Api-Key' auth header

ArthurZhou:
This PR fixes the Minimax authentication header by switching to 'X-Api-Key'. It incorporates and supersedes the changes from #117, as the original author restricted edits.


ArthurZhou:
fixed in #143

Issue #149, opened 2026-05-15
Models list empty for ChatGPT Codex provider and a few more things.


bnn1:
hey. there are a few issues i found:
1. after connecting to ChatGPT Codex provider, the `/models` command doesn't list any models, but you can manually enter model (e.g., `gpt-5.4`) and it seems to work.
2. entering `/config` command shows a menu with many options, but you can only navigate between some of them. consider this screenshot: 
<details>
<summary>screenshot</summary>

<img width="1140" height="586" alt="Image" src="https://github.com/user-attachments/assets/72c75d53-d3bc-4d1b-ba36-ccf0ccdda8c6" />
</details>
there's 'General' section but you can only navigate within 'Behaviour' section with arrow keys - no way to go into the General section. Same on all other tabs of config menu.  

3. you see a few hints on top of textarea: Ctrl+A, Ctrl+K and such, but the last hint that starts with `?` is out of viewport
<details>
<summary>screenshot</summary>

<img width="1694" height="704" alt="Image" src="https://github.com/user-attachments/assets/31269abf-7a33-40db-b55b-2842adc0b165" />
</details>

4. when cursor blinks, it shifts the whole prompt back and forth
<details>
<summary>gif</summary>

<img width="539" height="155" alt="Image" src="https://github.com/user-attachments/assets/e7955a95-8feb-4f06-bdd1-7f35bd6bf145" />
</details>

5. when you type in a long prompt, there's no line break in your prompt, it just goes and goes off-screen forever which makes it hard to read the prompt you write. i mean the whole text shifts with your input, and the beginning of your prompt goes offscreen.
6. tool calls results are unreadable - the text contrast is poor
<details>
<summary>screenshot</summary>

<img width="1411" height="327" alt="Image" src="https://github.com/user-attachments/assets/3cc7777b-d468-4788-a29e-8d09a4f8ecc1" />
</details>

That's all I found for now. I'll post updates if I find anything new. Thanks!

bnn1:
Tried `/managed-agents` command:
1. there's no auto-completion for this command
2. the command doesn't work. returns this:
<details>
<summary>screenshot</summary>

<img width="479" height="231" alt="Image" src="https://github.com/user-attachments/assets/41018c12-0a59-43bc-bd7b-121be9af49d4" />
</details>

bnn1:
A few other little things 😅.. 

[...]

PR #156, opened 2026-05-18
fix(tui): handle Shift with special characters on all terminals

jhult:
When terminals send unshifted characters with `SHIFT` modifier instead of the pre-shifted character (e.g., `SHIFT + 1` as `1` with `SHIFT` instead of `!`), the character insertion code was not applying the correct shift mapping. This occurred on terminals that do not apply shift transformations themselves, especially with the kitty keyboard protocol enabled.

Added `apply_shift_to_char()` helper to map numbers and special characters to their shifted equivalents (`1` → `!`, `2` → `@`, `3` → `#`, etc.). Updated the character insertion handler to apply this mapping when `SHIFT` modifier is present, alongside the existing uppercase mapping for ASCII letters.

Added 7 unit tests covering:
- Characters without `SHIFT` modifier pass through unchanged
- ASCII lowercase letters are uppercased with `SHIFT`
- All number-to-symbol mappings (`1` → `!`, `2` → `@`, etc.)
- All special character mappings (`-` → `_`, `=` → `+`, etc.)
- Already-shifted characters remain unchanged
- Other modifiers (`CTRL`, `ALT`) without `SHIFT` don't shift
- `SHIFT` combined with other modifiers still applies shift mapping

Also improved documentation to explicitly note US QWERTY limitation and explain why it's acceptable for non-US keyboard users.

Fixes part of Kuberwastaken/claurst#149

### DIFF
--- a/src-rust/crates/tui/src/app.rs
+++ b/src-rust/crates/tui/src/app.rs
@@ -581,6 +581,54 @@ fn layout_to_latin(c: char) -> String {
     mapped.unwrap_or(lower).to_string()
 }
 
+/// Apply shift transformation to a character based on standard US QWERTY layout.
+/// Handles both ASCII lowercase letters and number/symbol keys.
+///
+/// **Why this exists**: Terminals that support the kitty keyboard protocol send
+/// unshifted characters with modifier flags instead of pre-shifted characters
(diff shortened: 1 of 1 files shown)


Kuberwastaken:
Awesome LGTM

PR #159, opened 2026-05-19
Provider updates - fix auth and add Crof

jhult:
OpenCode Zen and other OpenAI-compatible providers are missing their environment variable mappings in `auth_store.api_key_for()`. While stored credentials are checked first, add comprehensive mappings as a safety fallback to ensure all providers can retrieve API keys from environment variables if `auth.json` is unavailable. Add [Crof.ai](https://crof.ai) as a new provider.


Kuberwastaken:
Awesome, much appreciated
Merging and will be looking for if we have other issues like this with other providers

PR #160, opened 2026-05-19
fix(tui): apply settings from config immediately on app startup

jhult:
Settings from `settings.json` (like show git repo, auto copy, etc.) were not taking effect until the user opened and closed the `/config` settings screen. The `SettingsScreen` was initialized with hardcoded default values instead of loading actual settings from disk on app startup.

Extract the settings application logic into a new `apply_settings_from_snapshot()` method and call it during `SettingsScreen` initialization. This ensures all saved settings are properly applied immediately when the app starts, without requiring the user to visit `/config` first. Also updated the test to handle the variable number of settings entries.


Kuberwastaken:
Awesome LGTM

PR #163, opened 2026-05-21
feat(tui): `@file` injection with typeahead autocomplete

jhult:
Implements end-to-end `@file` injection for the TUI prompt. Users type `@path` to reference files whose contents are automatically injected into the message before sending.

## Typeahead autocomplete

- Suggestions appear while typing `@path`, listing files and directories from cwd, absolute paths, or `~/` paths.
- Only directories with visible contents are shown (empty dirs or dirs that appear empty when show_hidden is off are suppressed).
- `Up`/`Down`/`Tab`/`Enter` navigate and accept suggestions; Enter also inserts a trailing space.
- Autocomplete is suppressed while the file injection dialog is open.
- Cursor-aware: `rfind('@')` scans only `text[..cursor]` so a mid-prompt `@` is handled correctly.

## File injection on submit
- `@refs` are resolved at submit time via `parse_at_refs`, which reads file contents and checks size limits.
- Directory refs are collected as `AtFileIssue::IsDirectory` and routed to a separate dismissible warning; they are never injected.
	image

- Oversized or binary files show a warning dialog offering "Enter to inject anyways" (bold red) or Esc to abort.
	image

- When both directories and file issues are present, the directory warning is shown first; on re-submit with force-inject, directory refs are silently dropped to prevent looping.

## Dialog and rendering
- `FileInjectionDialogState.oversized` now stores typed `AtFileIssue` values instead of serialized strings, eliminating the fragile string-contains pattern for `IsDirectory` detection.
- Dead path_relative_to helper removed; `is_directory_only()` uses `matches!` macro.
- Injected file contents are rendered as compact chips in the chat transcript rather than raw XML blocks.
- `@path` tokens in user message text are shortened to just `@filename` when chips are present.

### DIFF
--- a/src-rust/crates/cli/src/main.rs
+++ b/src-rust/crates/cli/src/main.rs
@@ -1947,7 +1947,7 @@ async fn run_interactive(
         app.notifications.tick();
 
         // Process file injection dialog outcome (if any)
-        if let Some((outcome, pending_input, _pending_imgs)) = app.file_injection_dialog.take_outcome() {
+        if let Some((outcome, pending_input, pending_imgs)) = app.file_injection_dialog.take_outcome() {
             use claurst_tui::FileInjectionOutcome;
 
             if matches!(outcome, FileInjectionOutcome::Abort) {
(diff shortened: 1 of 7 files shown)


Kuberwastaken:
Awesome

Issue #121, opened 2026-04-27
Deepseek need reasoning_content in thinking mode be passed back to API


linycm:
<img width="1629" height="112" alt="Image" src="https://github.com/user-attachments/assets/0e282482-6d26-437a-9e47-f7fc174adbd1" />

Deepseek need reasoning_content in thinking mode be passed back to API. could you revise for it?

PR #165, opened 2026-05-22
fix(deepseek): preserve reasoning_content for V4 thinking-mode multi-turn

jhult:
## Summary

Fixes DeepSeek V4 multi-turn tool-call failures by properly preserving `reasoning_content` through the streaming pipeline. DeepSeek V4's thinking mode requires reasoning_content to be echoed back on subsequent turns with tool calls, or the API returns a 400 error.

## Related

Based on [PR #111](https://github.com/Kuberwastaken/claurst/pull/111) which identified the issue and provided the fix approach.

## Changes

1. **openai_compat.rs stream loop**: Opens a dedicated Thinking block (using reserved index) on first reasoning delta and closes it before text/tool_calls/finish_reason. Ensures reasoning deltas are accumulated instead of silently dropped.

2. **Token optimization**: Only includes `reasoning_content` for providers that require it (currently DeepSeek V4). Added `requires_reasoning_roundtrip` quirk flag to gate inclusion, preventing wasted tokens on providers like OpenAI, Groq, Azure that ignore the field.

3. **Provider registration**: DeepSeek provider sets `requires_reasoning_roundtrip: true`.

## Testing

- ✅ Compiles without errors
- ✅ Block index strategy verified (uses usize::MAX - 100 to avoid collisions)
- ✅ Thinking lifecycle management correct (open on first delta, close before other content types)
- ✅ Backward compatible (field only included when needed)

## Notes

Research confirms this requirement is unique to DeepSeek V4; no other major LLM provider has multi-turn reasoning round-trip requirements.

### DIFF
--- a/src-rust/crates/api/src/providers/openai_compat.rs
+++ b/src-rust/crates/api/src/providers/openai_compat.rs
@@ -64,6 +64,12 @@ pub struct ProviderQuirks {
     /// reasoning output.  Example: `Some("reasoning_content")` for DeepSeek.
     pub reasoning_field: Option<String>,
 
+    /// Whether this provider requires reasoning_content to be echoed back on
+    /// subsequent turns in multi-turn conversations.  DeepSeek V4 is currently
+    /// the only provider with this requirement; most providers ignore this field.
+    /// When false, reasoning is not included in outbound messages to save tokens.
+    pub requires_reasoning_roundtrip: bool,
+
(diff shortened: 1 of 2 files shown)


jhult:
Fixes https://github.com/Kuberwastaken/claurst/issues/121

Kuberwastaken:
LGTM

PR #167, opened 2026-05-25
add self-hosted SearXNG web search backend

TheChuckster:
Brave search API is now $5 / month instead of free, no more free tier, add support for self-hosted SearXNG search


Kuberwastaken:
good idea! been using a version of this locally

PR #169, opened 2026-05-26
fix(tui): account for wide chars in prompt cursor

goodneamtakenbydogs:
## Summary

Fixes the prompt input cursor placement for Chinese/CJK wide characters in the TUI.

The prompt state stores the cursor as a UTF-8 byte offset, but rendering mapped that position to a visual column using character counts. In terminals, CJK characters commonly occupy two cells, so the overlay cursor could land one cell too far left and overwrite the following character.

## Changes

- Use `unicode-width` when mapping the prompt cursor byte offset to terminal display columns.
- Wrap prompt input lines by terminal cell width instead of Unicode scalar count.
- Add regression tests for cursor placement after a Chinese wide character.

## Validation

```bash
cargo test --package claurst-tui wide_character -j 2
cargo test --package claurst-tui prompt_input -j 2
```

The focused `prompt_input` suite passes locally: 111 passed.

### DIFF
--- a/src-rust/crates/tui/src/prompt_input.rs
+++ b/src-rust/crates/tui/src/prompt_input.rs
@@ -16,6 +16,7 @@ use ratatui::{
     text::{Line, Span},
     widgets::{Paragraph, Widget},
 };
+use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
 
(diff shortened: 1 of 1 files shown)


Kuberwastaken:
LGTM

PR #170, opened 2026-05-26
fix(api): respect Ollama base URL for native endpoints

goodneamtakenbydogs:
## Summary

Fixes remote Ollama configuration so the native Ollama endpoints follow the configured base URL.

When an Ollama provider is created, it keeps both:

- an OpenAI-compatible base URL used for chat completions, usually `<host>/v1`
- an Ollama native host used for `/api/tags` model discovery and health checks

`with_base_url()` updated the OpenAI-compatible base URL but left the native host pointing at the original `OLLAMA_HOST`/localhost value. That meant a remote `providers.ollama.api_base` could still leave model discovery or health checks targeting localhost.

## Changes

- Sync `ollama_native_host` when overriding an Ollama provider base URL.
- Strip a trailing `/v1` when deriving the native Ollama host.
- Add a regression test for a remote Ollama base URL override.
- Clarify the remote Ollama `api_base` behavior in the provider docs.

## Validation

```bash
cargo test --package claurst-api ollama_base_url_override_updates_native_host -j 2
cargo test --package claurst-api openai_compat -j 2
```

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -246,11 +246,13 @@ Native Cohere API adapter.
 
 Connects to a locally running Ollama instance. No API key required.
 
-**Base URL:** Reads `OLLAMA_HOST` (defaults to `http://localhost:11434`). Claurst appends `/v1` to construct the OpenAI-compatible endpoint.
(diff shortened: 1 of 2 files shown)


Kuberwastaken:
Thanks @goodneamtakenbydogs, the diagnosis was right. main fixed this independently in b097c7d (#192): `with_base_url` now retargets the Ollama (and LM Studio) native host, with tests, and docs/providers.md covers remote `api_base`. Closing as superseded; the branch also conflicts with main now. If you still see `/api/tags` hitting localhost with a remote `api_base` on current main, open an issue with your settings.json.

PR #171, opened 2026-05-26
chore(tui): remove unused mut in settings test

goodneamtakenbydogs:
## Summary

Removes an unnecessary `mut` from a settings screen unit test.

This clears the `unused_mut` warning emitted when compiling the `claurst-tui` test target.

## Validation

```bash
cargo test --package claurst-tui settings_screen --no-run -j 2
```

Before this change, the command reported 10 warnings including the `settings_screen.rs` `unused_mut` warning. After this change, that warning is gone and the remaining warnings are the existing non-snake-case test names in `prompt_input.rs`.

### DIFF
--- a/src-rust/crates/tui/src/settings_screen.rs
+++ b/src-rust/crates/tui/src/settings_screen.rs
@@ -825,7 +825,7 @@ mod tests {
 
     #[test]
     fn search_filters_entries_correctly() {
-        let mut screen = SettingsScreen::new();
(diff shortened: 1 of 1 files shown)


Kuberwastaken:
Thanks @goodneamtakenbydogs. This landed on main in 0b270d6 (clippy autofix sweep); `settings_screen.rs` already reads `let screen = SettingsScreen::new();` there, so this PR is now a no-op. Closing as superseded.

PR #174, opened 2026-05-29
Optimize file edit tools

MSIsunny:
1、Normalize Windows CRLF line endings so matching behavior is stable 
2、Enable batch edit to handle single file continuously


Kuberwastaken:
LGTM

Issue #185, opened 2026-06-19
[BUG] Claurst hangs when a provider stream pauses between tool-call chunks


N0zoM1z0:
## Summary

A deterministic OpenAI-compatible provider can start a streamed tool call and then pause before sending the tool arguments. Claurst does not return within the bounded run. The standalone reproducer downloads public `claurst v0.1.5`, starts a local mock provider, injects a long inter-chunk delay, and verifies the bounded hang without requiring this repository or any private fuzzing harness.

## Affected target

- Target: `claurst`
- Affected version checked with the public reproducer: `0.1.5`
- Current public latest checked on 2026-06-19: GitHub release `v0.1.5`
- Environment: Linux x86_64 Docker, public release asset
- Provider/model: OpenAI-compatible local mock provider, `gpt-4o-mini`

## Steps to reproduce

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It downloads the affected public Linux release asset, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/fac2ef03a6dead0a3e9c3939fe5be2b2

The Gist contains `claurst-hang-between-chunks.reproduce.py`. Download and run:

```bash
gh gist clone fac2ef03a6dead0a3e9c3939fe5be2b2 claurst-hang-between-chunks-reproducer
cd claurst-hang-between-chunks-reproducer
python3 claurst-hang-between-chunks.reproduce.py
```

To reuse an already-built local image:

```bash
python3 claurst-hang-between-chunks.reproduce.py --skip-build
```

Key output from a local run against the affected version:

```text
provider_requests_at_least: observed=1 expected>=1
timeout_or_signal: returncode=137 timed_out=True
REPRODUCED
```

## Actual behavior

Claurst sends one provider request, receives the start of a streamed shell tool call, and then remains blocked while the provider delays the next chunk. The reproducer prints `REPRODUCED` only when provider traffic is observed and the CLI fails to finish before the bounded timeout.

## Expected behavior

Claurst should apply a bounded stream/read timeout or recoverable cancellation path for incomplete tool-call streams, instead of hanging indefinitely when a provider stalls between chunks.

Kuberwastaken:
Will look into it, thank you!

[...]

Issue #175, opened 2026-05-29
[Feature Request] variable timeout


myway880:
I'm not sure if this is even possible or desirable but using a model on CPU could time out if the model is big.

On my PC some time it take ~8 minutes until first token when using Owen 3.6 35B A3B APEX MTP, by then claurst have already timed out and reset the connection .

Is it possible to add a user configurable timeout for users that are willing to wait for it

Kuberwastaken:
sure, I'll look into it!

myway880:
**Per-provider request timeout support + LM Studio server-side timeout limitation with large local models**

### Summary

When using large local models (e.g. Qwen3 30B–35B class) through **LM Studio** via the OpenAI-compatible endpoint, Claurst disconnects during prompt processing with the error:

> `[LM STUDIO SERVER] Client disconnected. Stopping generation...`

This happens even when setting very high values for `request_timeout_seconds`.

### Problem Description

Large local models can take a long time (sometimes several minutes) during the **prompt processing** phase before generating the first token. Claurst currently disconnects after roughly 40–60 seconds (sometimes up to ~5 minutes depending on conditions), making it unusable with bigger local models on LM Studio.

### Steps to Reproduce

1. Use a large model (30B+) via LM Studio.
2. Send a prompt that requires significant prompt processing time.
3. Claurst disconnects with the above LM Studio server log message.
4. Increasing `request_timeout_seconds` (even to 7200) has no effect.

### Investigation & Findings

- Added support for `request_timeout_seconds` per provider + global `default_request_timeout_seconds`.
- The setting is correctly read and applied to the `reqwest` client.
- Setting the value to `0` now correctly disables the timeout (no longer causes immediate connection failure).
- However, the disconnection still occurs because **LM Studio's server has its own hardcoded timeout** (reported ~300 seconds in multiple versions) on the OpenAI-compatible endpoint.

This is a known limitation on the LM Studio side (see related reports in lmstudio-bug-tracker and other tools like Cline/Roo Code).

Note: During my attempt to compile this i discovered that the release tag 1.4 zipped code is missing these files: spinner.rs and account.rs, so if you want to port this to 1.4 you will need to copy these files from main back to 1.4 to compile successfully 
### Changes Made (Patch)

I implemented the following improvements in Claurst:

**Files modified:**
- `crates/core/src/lib.rs`
  - Added `request_timeout_seconds` field to `ProviderConfig`
  - Added `default_request_timeout_seconds` to `Config`
[...]

PR #187, opened 2026-06-21
variable time out

myway880:
**Per-provider request timeout support + LM Studio server-side timeout limitation with large local models**

### Summary

When using large local models (e.g. Qwen3 30B–35B class) through **LM Studio** via the OpenAI-compatible endpoint, Claurst disconnects during prompt processing with the error:

> `[LM STUDIO SERVER] Client disconnected. Stopping generation...`

This happens even when setting very high values for `request_timeout_seconds`.

### Problem Description

Large local models can take a long time (sometimes several minutes) during the **prompt processing** phase before generating the first token. Claurst currently disconnects after roughly 40–60 seconds (sometimes up to ~5 minutes depending on conditions), making it unusable with bigger local models on LM Studio.

### Steps to Reproduce

1. Use a large model (30B+) via LM Studio.
2. Send a prompt that requires significant prompt processing time.
3. Claurst disconnects with the above LM Studio server log message.
4. Increasing `request_timeout_seconds` (even to 7200) has no effect.

### Investigation & Findings

- Added support for `request_timeout_seconds` per provider + global `default_request_timeout_seconds`.
- The setting is correctly read and applied to the `reqwest` client.
- Setting the value to `0` now correctly disables the timeout (no longer causes immediate connection failure).
- However, the disconnection still occurs because **LM Studio's server has its own hardcoded timeout** (reported ~300 seconds in multiple versions) on the OpenAI-compatible endpoint.

This is a known limitation on the LM Studio side (see related reports in lmstudio-bug-tracker and other tools like Cline/Roo Code).

Note: During my attempt to compile this i discovered that the release tag 1.4 zipped code is missing these files: spinner.rs and account.rs, so if you want to port this to 1.4 you will need to copy these files from main back to 1.4 to compile successfully 
### Changes Made (Patch)

I implemented the following improvements in Claurst:

**Files modified:**
- `crates/core/src/lib.rs`
  - Added `request_timeout_seconds` field to `ProviderConfig`
  - Added `default_request_timeout_seconds` to `Config`
  - Added `resolved_request_timeout()` helper method
- `crates/api/src/providers/openai_compat.rs`
  - Added `with_request_timeout(Duration)` method
  - Made `0` mean "no timeout" (skip setting `.timeout()`)
- `crates/api/src/providers/openai.rs`
  - Same `with_request_timeout()` improvement
- `crates/api/src/registry.rs`
  - Resolve and apply per-provider timeout when constructing providers

**Key behavior:**
- `request_timeout_seconds: 0` → Disables HTTP timeout completely (useful for slow local models)
- Positive values work as expected
- Local providers can now be configured independently of cloud providers

### Current Status

- The client-side timeout feature is implemented and working.
- The root cause of the disconnection with LM Studio is on the **server side** (LM Studio's internal timeout), not in Claurst.
- Even with timeout disabled in Claurst, LM Studio still closes the connection after its internal limit.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+#
+# compile.sh - Build / Check the Claurst Rust workspace
+#
+# Usage:
+#   ./compile.sh            # Fast check (recommended during development)
+#   ./compile.sh release    # Full optimized build
+#   ./compile.sh test       # Run tests
+#
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SRC_RUST_DIR="$SCRIPT_DIR/src-rust"
+
+cd "$SRC_RUST_DIR"
+
+echo "==> Working directory: $(pwd)"
+echo "==> Rust version: $(rustc --version)"
+echo ""
+
+case "${1:-check}" in
+    check|"")
+        echo "==> Running: cargo check --workspace"
+        cargo check --workspace
+        echo ""
+        echo "✅ Check passed successfully."
+        ;;
+
+    build|release)
+        echo "==> Running: cargo build --release --workspace"
+        cargo build --release --workspace
+        echo ""
+        echo "✅ Release build completed."
+        echo "   Binaries are in: target/release/"
+        ;;
+
+    test)
+        echo "==> Running: cargo test --workspace"
+        cargo test --workspace
+        echo ""
+        echo "✅ All tests passed."
+        ;;
+
+    clippy)
+        echo "==> Running: cargo clippy --workspace --all-targets -- -D warnings"
+        cargo clippy --workspace --all-targets -- -D warnings
+        echo ""
+        echo "✅ Clippy passed with no warnings."
+        ;;
+
+    fmt)
+        echo "==> Running: cargo fmt --all -- --check"
+        cargo fmt --all -- --check
+        echo ""
+        echo "✅ Formatting is correct."
+        ;;
+
+    *)
+        echo "Unknown command: $1"
+        echo ""
+        echo "Usage:"
+        echo "  ./compile.sh            # Fast check (default)"
+        echo "  ./compile.sh release    # Full release build"
+        echo "  ./compile.sh test       # Run all tests"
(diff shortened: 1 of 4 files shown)


Kuberwastaken:
Thank you for the genuinely excellent investigation here, @myway880 — the repro, the LM Studio server-log detail, and tracing it to prompt-processing time before first token were all spot on, and they directly informed how this got solved on \`main\`.

Since this PR was opened, per-provider **and** global request timeouts landed on \`main\` (via the timeout work in #175/#185), and the result is a superset of what this PR proposed:

- **Per-provider override:** `ProviderConfig::requestTimeoutSecs` (accepts the snake_case alias `request_timeout_secs` too) — set it just for LM Studio without touching cloud providers.
- **Global default:** `Config.request_timeout_secs`.
- **Precedence:** per-provider → global → `DEFAULT_REQUEST_TIMEOUT_SECS` (now **600s**, up from the old ~40–60s that was killing you).
- **Applied to the reqwest client** for both `openai` and `openai_compat` providers, resolved in `registry.rs`.
- **Progress-aware idle timeout (#185):** the streaming loop bounds *mid-stream stalls* but the idle window is never smaller than your configured request timeout **and resets on every chunk that makes progress** — so a large local model that spends minutes on prompt processing (but is genuinely working) is no longer disconnected. This was the core of your report.

Example `~/.config/claurst/settings.json`:
```jsonc
{
  "providers": {
    "lmstudio": {
      "requestTimeoutSecs": 7200   // or any high value for big local models
    }
  }
}
```

On the two design points from your patch:
- **`0` handling:** on `main`, `0` intentionally falls back to the 600s default rather than removing the timeout entirely — we opted against a truly-unbounded client timeout (it can wedge a session on a genuinely dead connection). For effectively-unlimited local runs, set a very large value (e.g. `86400`). If a real no-cap need comes up we can revisit.
- **LM Studio’s own ~300s server-side cap:** as you found, that one lives on the LM Studio side and is outside claurst’s control — nothing we can fix from here.

Because the feature already exists on `main` in a more complete form (and this branch now conflicts), I’m closing this out as **addressed** — but the fix is very much thanks to your report. Really appreciate you digging in and writing it all up. 🙏

PR #189, opened 2026-06-26
fix(tools): stop bash tool hanging on detached child processes

### LINKED ISSUE #184: [BUG] Claurst hangs after interrupted child-process tool call
## Summary

When Claurst receives a tool call that starts a child process and the CLI is interrupted during the tool turn, the process does not complete cleanly within the bounded run. The linked standalone reproducer downloads public `claurst v0.1.5`, starts a local mock provider, sends a child-spawning shell tool call, interrupts the CLI, and verifies the behavior without requiring this repository or any private fuzzing harness.

## Affected target

- Target: `claurst`
- Affected version checked with the public reproducer: `0.1.5`
- Current public latest checked on 2026-06-18: GitHub release `v0.1.5`
- Environment: Linux x86_64 Docker, public release asset
- Provider/model: OpenAI-compatible local mock provider, `gpt-4o-mini`

## Steps to reproduce

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It downloads the affected public Linux release asset, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/85df17ed025664d223f93de548c10e0a

The Gist contains `claurst-child-process-cleanup.reproduce.py`. Download and run:

```bash
gh gist clone 85df17ed025664d223f93de548c10e0a claurst-child-process-cleanup-reproducer
cd claurst-child-process-cleanup-reproducer
python3 claurst-child-process-cleanup.reproduce.py
```

To reuse an already-built local image:

```bash
python3 claurst-child-process-cleanup.reproduce.py --skip-build
```

Key output from a local run against the affected version:

```text
provider_requests_at_least: observed=1 expected>=1
side_effect_file_exists interrupt-child.pid: bytes=... expected>=1
timeout_or_signal: returncode=... timed_out=True
REPRODUCED
```

## Actual behavior

Claurst accepts an interrupted child-spawning tool call, records the PID side effect, and then fails to finish normally before the bounded timeout. The reproducer prints `REPRODUCED` only when the provider request is observed, the child-spawning command records its PID, and the CLI exits by timeout or signal.

## Expected behavior

Claurst should track, terminate, and reap tool-started child processes across interruption, then return control without hanging.

Ijtihed:
## What

Fixes #184. The bash tool's foreground path read the child's stdout to
EOF before calling `wait()`. A tool-spawned process that detaches into
its own session (`setsid` / `start_new_session`) inherits the pipe and
holds the write end open, so EOF never arrives -- the tool blocks long
after the command finished and the agent turn hangs until the timeout
or an interrupt signal.

## Fix

- Read stdout/stderr concurrently with `wait()` and stop as soon as the
  direct child exits; drain only the already-buffered output instead of
  waiting for pipe EOF a detached grandchild can withhold. This also
  removes the old sequential stdout-then-stderr read, which could
  deadlock when a command filled the stderr pipe buffer.
- Add `kill_on_drop(true)` to the foreground and Windows spawns so an
  interrupted turn kills the child instead of leaking it. The background
  path already got this in 5a0b501; these two were missed.

## Test plan

Added `foreground_does_not_hang_on_detached_grandchild` (Unix): runs
`setsid sleep 30 & echo spawned` and asserts `execute` returns promptly.

- Before: FAILED -- `execute hung (regression of #184): Elapsed(())`,
  hit the 10s guard.
- After: ok in 0.11s. Full `claurst-tools` suite: 77 passed, 0 failed.

Also verified manually (not in the committed tests): a 5000-line stdout
command returns untruncated, and a command writing ~40KB to stderr no
longer deadlocks.

### DIFF
--- a/src-rust/crates/tools/src/bash.rs
+++ b/src-rust/crates/tools/src/bash.rs
@@ -20,6 +20,95 @@ use tracing::debug;
 /// this marker is metadata (final pwd + env dump) rather than user-visible output.
 const SHELL_STATE_SENTINEL: &str = "__CC_SHELL_STATE__";
 
+/// Read one line from an optional line reader, mapping read errors / EOF to
+/// `None`. Used by the `tokio::select!` loop in [`drive_child`].
+async fn next_line<R>(reader: &mut Option<tokio::io::Lines<R>>) -> Option<String>
+where
+    R: tokio::io::AsyncBufRead + Unpin,
+{
+    match reader {
+        Some(lines) => lines.next_line().await.ok().flatten(),
+        None => None,
+    }
+}
+
+/// Pull any *already-buffered* lines from `reader` into `sink`, stopping at EOF,
+/// a read error, or after a short grace period with no new data.
+///
+/// Called after the direct child has exited: it must never block on pipe EOF,
+/// because a tool-spawned grandchild that detached into its own session can keep
+/// the pipe's write end open indefinitely (issue #184).
+async fn drain_buffered<R>(reader: &mut Option<tokio::io::Lines<R>>, sink: &mut Vec<String>)
+where
+    R: tokio::io::AsyncBufRead + Unpin,
+{
+    let Some(lines) = reader else { return };
+    // Loop ends on the first non-`Ok(Ok(Some))`: EOF, a read error, or no data
+    // within the grace window (a detached grandchild holding the pipe open).
+    while let Ok(Ok(Some(line))) =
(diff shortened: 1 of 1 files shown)


Kuberwastaken:
LGTM king

PR #194, opened 2026-06-29
fix(models): surface latest Claude models (Opus 4.8) in provider pickers

### LINKED ISSUE #180: Unable to bump to latest Claude model
I am unable to bump claude version to Opus 4.8. It says on 4.6 regardless. Tried both ways:
* /connect
* /model

Am I missing anything something?

Kuberwastaken:
## Problem

Users were stuck on Claude Opus 4.6 and could not select Opus 4.8 via `/connect` or `/model`, even though 4.8 is the current flagship.

## Root cause

The primary blocker was **not** the bundled snapshot. When the model picker opens, it first fills the list from the registry, then spawns a live `provider.list_models()` call and **overwrites** the list with that result (`crates/cli/src/main.rs` ~3360). The Anthropic (and Bedrock) `list_models()` returned a hardcoded curated list whose newest Opus was `claude-opus-4-6`, so 4.8 never showed up regardless of what the snapshot/cache contained.

Secondary: the bundled `models-snapshot.json` predated Opus 4.8, and the hardcoded `DEFAULT_MODEL`/`OPUS_MODEL` constants still pointed at 4.6.

## Changes

- `crates/api/src/providers/anthropic.rs` — add `claude-opus-4-8` as the first Opus entry in `list_models()` (the curated list the picker actually displays). Kept 4.6 so existing users aren't forced off.
- `crates/api/src/providers/bedrock.rs` — parallel fix: add `anthropic.claude-opus-4-8` to the Bedrock curated list.
- `crates/api/assets/models-snapshot.json` — refreshed via `scripts/sync-models.sh` (network to models.dev was available). Now includes Opus 4.7/4.8 and current pricing/context (145 providers).
- `crates/core/src/lib.rs` — bump `DEFAULT_MODEL` and `OPUS_MODEL` from `claude-opus-4-6` to `claude-opus-4-8`; bump the `anthropic/claude-opus-4-6` manager model in the `anthropic-tiered` and `cross-opus-flash` managed-agent presets to `claude-opus-4-8` (and the matching description text).

## Notes / scope decisions

- **TUI background refresh (was listed as a fix item) is already wired.** `run_interactive` (the only production TUI launch path) already calls `spawn_models_cache_refresh()` before the event loop (`crates/cli/src/main.rs` ~1729, since v0.0.9), and the picker reloads the cache on open. So no new refresh code was needed; the actual gap was the hardcoded `list_models()` overwrite described above.
- **Bedrock default pin** (`anthropic.claude-sonnet-4-6-v1` in `Config::effective_model`) was left as-is: it is a Sonnet-tier default (intentionally different from the Anthropic Opus default), and the latest Sonnet is still 4-6.
- `SONNET_MODEL`/`HAIKU_MODEL` constants are already current (4-6 / 4-5) per the refreshed snapshot — left unchanged.
- The default-model resolver (`best_model_for_provider`) is registry-driven and now resolves Anthropic to `claude-opus-4-8` (newest `claude-opus-4` by release date) from the refreshed snapshot.

## Verification

- `cargo check -p claurst` — clean.
- `cargo test -p claurst-api -p claurst-tui -p claurst-core` — all pass (0 failures); existing `model_registry` tests still green.
- Snapshot validated as JSON; single-line format and no trailing newline preserved (diff is a clean 1-line swap).

Closes #180


Kuberwastaken:
Superseded by #200, which fixes #180 the clean (opencode-style) way: the models.dev catalog is the single source of truth, the picker projects from it instead of being overwritten by a provider's hardcoded list, and provider-id canonicalization (qwen/alibaba) is added. #200 also carries the Opus 4.8 snapshot entry, so the hardcoded-list edits here are no longer needed. Closing in favor of #200.

PR #308, opened 2026-07-09
fix(api): refresh MiniMax M3 metadata

octo-patch:
## Summary
- Refresh the bundled models.dev snapshot so MiniMax-M3 is available with current context, modality, thinking, and pricing metadata while MiniMax-M2.7 remains available.
- Keep the Anthropic-compatible base at `/anthropic` because Claurst appends `/v1/messages`, and add regression coverage for the final URL and regional overrides.
- Translate MiniMax-M3 thinking requests to the provider's `adaptive` schema while leaving M2.7's always-on thinking implicit.
- Preserve generic context pricing tiers from models.dev and document the global/China Anthropic and OpenAI-compatible entry points.

## Test plan
- [x] `cargo test -p claurst-api` (100 passed)
- [x] `cargo test -p claurst-core` (521 passed across unit and integration tests)
- [x] `cargo test -p claurst-query` (138 passed)
- [x] `jq empty src-rust/crates/api/assets/models-snapshot.json`
- [x] MiniMax target metadata and current models.dev provider-node consistency checks
- [x] `git diff --check`


octo-patch:
Updated the PR with a follow-up commit. Because Claurst appends `/v1/messages`, the MiniMax base now remains `/anthropic`, producing exactly one `/v1/messages`; I also aligned M3 adaptive thinking, retained the 512k pricing tier, and documented regional endpoints. I ran `cargo test -p claurst-api`, `cargo test -p claurst-core`, `cargo test -p claurst-query`, and `git diff --check`.

Kuberwastaken:
awesome lgtm

octo-patch:
Correction: #308 merged before the final follow-up commit reached its PR head, so the merged diff ends at `fd70054`. The remaining verified MiniMax catalog scoping, API-root normalization, priority-tier support, pricing documentation, and request-capture coverage are now in https://github.com/Kuberwastaken/claurst/pull/313.

Issue #314, opened 2026-07-14
caching is not working for local models


myway880:
the caching counter is always at 0k/262k and it never goes up. (I'm assuming this the model context cache)
edit this maybe a problem in the model i'm using and how claurst manage its thinking process it seems the model keep spending all its token budget on thinking and it never does anything, so feel free to close this I'm not really sure what the problem is at the moment
the fact that i have to repeat myself seem to be due to the fact that pasted text is not expanded and it is seen as empty 

`old text:
every prompt start from the beginning with no context and the AI is usually very confused about what is the requirement i have to repeat myself. 
As a workaround? I had to enable caching on the server side. but I think that it is a very small cache because even though the model seem to knows what the requirement are still keeps starting the thinking from the beginning every time. `

`I'm not sure if it is related but i had a hard time making it create any new files or change the existing files, it just goes on for hours running in circles thinking but never really doing any changes .
`
note: I have not used claurst with any non local so this could be a general problem, or it could be something in my setup 
Here are the server setting i used running the models in case the issue is not in claurst , 
```
llama-server.exe -m E:\Models\Qwen3.6-35B-A3B-APEX-MTP-Quality.gguf --host 127.0.0.1 --port 8080 --no-webui --jinja --ctx-size 70697 --n-gpu-layers 12 --n-cpu-moe 0 --main-gpu 0 --tensor-split 0 --split-mode layer --batch-size 2048 --ubatch-size 1024 --threads 6 --parallel 4 --cache-type-k f16 --cache-type-v f16 --flash-attn on --kv-offload --kv-unified --no-direct-io --mlock --spec-type draft-mtp --spec-draft-n-max 2 --spec-draft-n-min 0 --spec-draft-p-min 0.75 --timeout 0 --cache-prompt --no-context-shift --ctx-checkpoints 64 --checkpoint-min-step 512
```
```
llama-server.exe -m E:\Models\gemma-4-26B-A4B-it-QAT-Q4_0.gguf --port 8080 -c 72400 -ngl 11 -t 6 --kv-unified -np 4 -fa on --mlock --timeout 0 -b 2048 --ubatch-size 1024 --jinja  --cache-prompt --no-context-shift --ctx-checkpoints 64 --checkpoint-min-step 512
```
can you please create some documentation about how to use locally, like what are the recommended server cli commands or settings and what are the best models to use ?
Note 1: I also tried running claurst with CLAURST_COORDINATOR_MODE=1 and without it  
[...]

PR #316, opened 2026-07-16
fix(tui): paste pipeline — stop truncating pastes in the live event loop; click a placeholder to read it

Kuberwastaken:
## Problem

Large pastes kept arriving truncated or empty at the model (reported again in #314: *"pasted text is not expanded and it is seen as empty"*), even after the earlier placeholder-expansion fix (45a775c).

**Root cause:** every paste fix so far went into `App::run` in `tui/app.rs` — an event loop **nothing calls**. The real binary runs `run_interactive` in `cli/main.rs`, which has its own loop that:

1. had **no paste-burst detection** — on terminals without bracketed paste (Windows Ctrl+V, some tmux setups) the clipboard floods in as raw key events, every pasted newline acts as Enter, and a truncated first line gets submitted while the rest sprays into follow-up messages;
2. **silently dropped `Event::Paste` during streaming**, even though the prompt stays editable to compose queued follow-ups;
3. bypassed `handle_paste_data`, skipping file-path/image paste detection.

## Fix

- Run the burst drain in the live loop *before* Enter/submit handling; replay the drain's leftover key as a synthetic event.
- Repair the drain itself: skip Windows Press/Release Release events (they previously capped every burst at one char), map Ctrl+J (raw LF) back to a newline, push Enter as `\r` so `normalize_newlines` collapses CRLF pairs, and keep literal tabs.
- Deliver bracketed pastes while streaming, routed through the shared `handle_paste_data` path.
- **Click a `[Pasted text #N +X lines]` placeholder to read it**: opens a read-only scrollable modal (wheel/arrows/PgUp/PgDn/g/G, Esc closes) instead of destructively splicing hundreds of lines into a 3-row prompt. Alt+E — from the prompt or inside the viewer — still expands in place for editing.

## Verification

- `cargo test -p claurst-tui`: 657 passed (new viewer/click tests included; click-to-expand tests updated to click-to-view).
- Manual, against the built binary in tmux: raw 4-line key flood → one placeholder, zero premature submits, line breaks intact; bracketed paste → placeholder; click → viewer shows full body; Esc keeps placeholder; Alt+E expands in place.
- Submit-time expansion (placeholder → real body) covered by existing `take()` unit tests.

Fixes the paste-truncation half of #314 (the cache-counter half is addressed separately).

### DIFF
--- a/src-rust/crates/cli/src/main.rs
+++ b/src-rust/crates/cli/src/main.rs
@@ -2187,7 +2187,11 @@ async fn run_interactive(
         // Poll for crossterm events (keyboard/mouse) with short timeout
         // unless an auto-submit (queued message) is pending — in which case
         // synthesize an Enter event to dequeue and submit it.
-        let synthetic_event: Option<Event> = if app.pending_auto_submit && !app.is_streaming {
+        let synthetic_event: Option<Event> = if let Some(k) = app.pending_key.take() {
+            // A non-character key swallowed by the paste-burst drain — replay
+            // it so the keystroke that ended a raw-key paste is not lost.
+            Some(Event::Key(k))
+        } else if app.pending_auto_submit && !app.is_streaming {
             app.pending_auto_submit = false;
             Some(Event::Key(crossterm::event::KeyEvent::new(
                 KeyCode::Enter,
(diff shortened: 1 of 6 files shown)


Kuberwastaken (commented):
## Review

**What it does:** Moves the paste pipeline fixes into the event loop the binary actually runs (`run_interactive`), fixes the burst drain itself, and adds a read-only viewer so a `[Pasted text #N]` placeholder can be read by clicking it.

### Correctness
- ✅ Root cause is real and verified: `App::run` (where all prior paste handling lived) has no callers; the live loop had no burst detection and dropped `Event::Paste` mid-stream.
- ✅ Burst false-positives are benign: 2+ chars landing in the same 16 ms poll window from fast typing are inserted verbatim via `prompt_input.paste()` (identical outcome to typing); `detect_pasted_path` can't match non-absolute fragments.
- ✅ Replay ordering: `pending_key` takes priority over `pending_auto_submit`, which stays set for the next iteration — nothing is lost.
- ✅ Submit while viewer open is impossible (`any_modal_open()` includes the viewer, and it swallows keys in `handle_key_event`).
- ✅ Placeholder/body survive view-then-Esc; Alt+E from the viewer no-ops gracefully if the placeholder was edited away (`expand_paste_ref_by_id` re-parses).
- ⚠️ **Windows CRLF caveat:** the drain maps Enter→`\r` and Ctrl+J→`\n` so `normalize_newlines()` collapses CRLF pairs. If some Windows console path delivers *both* CR and LF as `Enter` key events, blank doubled lines could appear. Strictly better than today's submit-per-newline, but worth a follow-up test on a real Windows box.

### Tests
- 657 tui unit tests pass, including new viewer state/wrap/clamp tests and the rewritten click tests (click-to-expand → click-to-view, Alt+E round-trip).
- The drain itself isn't unit-testable (reads the global crossterm queue) — covered by manual tmux verification (raw flood → one placeholder, no submits, breaks intact; bracketed paste unaffected).

### Style/quality
- Viewer follows house modal conventions (`begin_modal_buf`, `last_max_scroll` Cell-clamp pattern, diff_viewer-style key handling).
- One nit found and fixed in 2b43e58 (duplicated doc comment on `pending_key`).

Merging once CI is green on all three platforms.

PR #317, opened 2026-07-16
fix(tui): reclaim fixed chrome — scroll welcome banner away, compact footer (#310)

### LINKED ISSUE #310: UI: Reduce fixed header/footer height to reclaim content space
Image

### Problem
The top banner in the TUI is fixed and cannot scroll, occupying significant vertical space. In Claude Code, the equivalent top area scrolls with the content, freeing up room for the conversation. Similarly, the bottom footer in Claurst is noticeably larger than Claude Code's, further squeezing the content area.
### Expectation
Reduce the fixed UI chrome so that the main conversation content area gets more vertical space — ideally matching Claude Code's more compact header and footer layout.

Kuberwastaken:
## Problem

The two-column welcome box was pinned as a **fixed 9-row header** at the top of the message pane for the entire session — not just on the welcome screen but throughout an active conversation. The footer reserved **2 rows** but drew content in only the first (the second was blank padding). Together this chrome permanently consumed ~11 rows, so on an 80x24 terminal the conversation transcript was squeezed into ~8 rows. Claude Code, by contrast, lets its banner scroll away with the conversation and keeps a 1-line footer.

## Approach

Approach (a) from the issue — render the welcome banner as ordinary transcript content so it scrolls away — with a compact footer:

- **Welcome box is now a welcome _screen_, not a header.** The rich two-column box (mascot, tips, recent activity) is rendered only while the transcript is empty. Startup notices (remote session, `+dir`, away summary) sit just below it there, exactly as before.
- **During a conversation, a compact welcome banner is prepended to the transcript** as ordinary scrollable content (in `build_all_items` and the streaming committed-prefix path in `render_streaming_items`). It carries the greeting the box led with plus a getting-started hint (`/help for commands  ·  ? for shortcuts`) and any startup notices, in ~4 rows inside a rounded box. Because it is part of the message list, it scrolls away with the conversation and leaves **zero fixed header rows**.
  - The banner is deliberately IO-free — it does **not** call `select_tip` (which reads the tip history from disk) — so it is cheap to rebuild every streaming frame and is byte-identical between the full-rebuild and cached-prefix render paths (the streaming-cache equivalence tests still pass).
- **Footer shrunk from 2 rows to 1.**
- Mouse-click geometry is unaffected: `last_msg_area` / `last_input_area` are still set from the real rects, and the error-modal anchor (`footer_right_column_area`) is cleared when the welcome box is not on screen so modals fall back to a sane placement.

## Before / after row-count accounting (80x24, active conversation)

| Region | Before | After |
| --- | --- | --- |
| Fixed welcome-box header | 9 rows (+ notice rows) | 0 rows (compact banner is scrollable, ~4 rows, scrolls away) |
| Footer | 2 rows (1 content + 1 blank pad) | 1 row |
| **Total fixed chrome** | **11 rows** | **1 row** |
| **Conversation viewport** | **~8 rows** | **~18 rows** (+125%) |

Verified by `tmux` capture at 80x24: before, the 9-row box stayed pinned above the first message; after, the compact banner sits at the top of the transcript with the message directly beneath it and scrolls off once content fills the viewport (confirmed at 40x12, where only the tail remains visible). No panics at 40x12 or 22x10 (the latter exercises the `width < 24` single-line banner fallback). The empty welcome screen is unchanged.

## Verification

- `cargo test -p claurst-tui --lib` — 652 passed, 0 failed (includes the streaming-cache equivalence tests proving `banner ++ prefix ++ tail == build_all_items`, and an updated render smoke test asserting the banner scrolls with content and the full box's "Recent activity" panel is absent mid-conversation).
- `cargo clippy -p claurst-tui --all-targets` — no new warnings.
- Manual `tmux` testing at 80x24, 40x12, 22x10.

Closes #310

### DIFF
--- a/src-rust/crates/tui/src/lib.rs
+++ b/src-rust/crates/tui/src/lib.rs
@@ -840,7 +840,11 @@ mod tests {
     }
 
     #[test]
-    fn test_render_app_keeps_logo_header_after_first_message() {
+    fn test_render_app_shows_scrollable_banner_after_first_message() {
+        // Once a conversation starts, the welcome box is no longer pinned as a
+        // fixed header (issue #310): a compact banner leads the transcript (and
+        // scrolls away with content), while the rich two-column welcome box — its
+        // mascot and "Recent activity" panel — is only the empty-screen welcome.
         let backend = TestBackend::new(120, 40);
         let mut terminal = Terminal::new(backend).unwrap();
         let mut app = make_app();
@@ -859,8 +863,13 @@ mod tests {
             .collect::<Vec<_>>()
             .join("");
 
+        // The compact banner (title + hint) and the message both render.
         assert!(rendered.contains("Claurst"));
+        assert!(rendered.contains("? for shortcuts"));
         assert!(rendered.contains("hello"));
+        // The full welcome box's recent-activity panel must NOT be drawn during
+        // a conversation — that would mean the old fixed header is still there.
+        assert!(!rendered.contains("Recent activity"));
     }
 
     #[test]
--- a/src-rust/crates/tui/src/render.rs
+++ b/src-rust/crates/tui/src/render.rs
@@ -598,7 +598,7 @@ pub fn render_app(frame: &mut Frame, app: &App) {
             Constraint::Length(status_height),
             Constraint::Length(prompt_height),
             Constraint::Length(suggestions_height),
-            Constraint::Length(2),
+            Constraint::Length(1),
         ])
         .split(size);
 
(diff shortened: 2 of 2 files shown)


Kuberwastaken (commented):
## Review

**Approach** — took option (a) from the issue: the rich two-column welcome box becomes a true welcome *screen* (empty transcript only), and during conversations a compact ~4-row banner is prepended to the transcript as ordinary scrollable content. Fixed chrome goes from ~11 rows (9 header + 2 footer) to 1 footer row — on an 80x24 terminal the transcript roughly doubles.

- ✅ The subtle risk here was the streaming committed-prefix cache, and it's handled correctly: the banner is prepended *outside* the cached prefix in both render paths, so cache entries stay byte-identical between `build_all_items` and `render_streaming_items` (equivalence tests still pass).
- ✅ Banner is deliberately IO-free (no `select_tip` disk read) — safe to rebuild every streaming frame.
- ✅ Banner rows are pushed with no message index, so `message_row_map` click/selection mapping shifts consistently in both paths.
- ✅ `footer_right_column_area` is explicitly cleared once the welcome box leaves the screen — better than the stale anchor the old code left behind on small terminals.
- ✅ Behavior test updated to assert the new contract (banner + no 'Recent activity' panel mid-conversation), not just tweaked to pass.
- ✅ CI green on all 3 platforms; 2 scoped commits, clean messages.

**Non-blocking:** `USER`/`USERNAME` env lookup runs per banner rebuild (every streaming frame) — trivially cheap, but could be memoized if it ever shows up in a profile.

Note: base predates #316/#318 merges; hunks don't overlap either (checked), so merging directly.

PR #318, opened 2026-07-16
feat: support local model metadata overrides (context window, etc.)

### LINKED ISSUE #309: Feature: Support local model metadata override (e.g. context window size)
### **Background & Motivation**
When users connect to self-hosted services via OpenAI Compatible, or use model names (aliases) that are not listed on models.dev, models.dev may incorrectly match the alias to a different model. This results in inaccurate metadata (e.g. context window size, max output tokens) being displayed in the TUI model picker.

### **Expectation**
Claurst should provide a local mechanism for users to supply or correct model metadata (especially context window size), so the model picker and related UI can display accurate information without relying entirely on models.dev.

Kuberwastaken:
## Problem

When users connect self-hosted endpoints via the OpenAI-compatible provider
(or use model aliases models.dev does not know), claurst's metadata lookup can
mismatch the alias to an unrelated models.dev entry — or find nothing at all —
so the model picker, token-usage warnings, and auto-compact thresholds size the
model with the wrong context window / max output tokens.

Closes #309.

## Approach

Add a local, user-editable metadata override that layers cleanly:
**user override > models.dev > built-in defaults.**

- **`ModelOverride` config type** (`crates/core`): optional `contextWindow`,
  `maxOutputTokens`, `name`, `releaseDate`, `status`. Keyed by
  `"provider/model"` in a `modelOverrides` map on both the top level of
  `settings.json` and the nested `config` block (mirrors the existing
  `providers` / `provider_configs` convention). Field names accept camelCase or
  snake_case. `Settings::effective_config()` and `Settings::merge()` fold the map
  in.

- **Registry layering** (`crates/api`): `ModelRegistry` retains the overrides and
  applies them over the catalog — existing entries are patched in place, and any
  `"provider/model"` key the catalog lacks (a self-hosted alias) is materialised
  into a synthetic entry. Because every context-window consumer funnels through
  `ModelRegistry::get()` / `list_*_by_provider()`
  (`App::refresh_context_window_size`, `compact::resolve_context_window` for
  auto-compact + token warnings, and the picker's catalog projection), correcting
  the registry corrects all of them at once. Overrides are re-asserted after
  every `load_cache()` and `refresh_from_models_dev()` so a later catalog merge
  can't clobber them.

- **Wiring**: overrides are applied in `App::new`, `App::apply_provider_refresh`,
  and the CLI's `load_cached_model_registry`. The live-discovery picker path
  (self-hosted / openai-compatible endpoints) also honours the override when
  rendering the discovered list.

## Config syntax

Add to `~/.claurst/settings.json` (top level, or under `config`), keyed by the
fully-qualified `provider/model` id:

```json
{
  "modelOverrides": {
    "custom-openai/my-local-llm": {
      "contextWindow": 32768,
      "maxOutputTokens": 4096,
      "name": "My Local LLM",
      "releaseDate": "2026-01-01",
      "status": "beta"
    },
    "ollama/qwen3-coder-30b": { "contextWindow": 262144 }
  }
}
```

All fields are optional (an unset field keeps the catalog value); field names
accept camelCase (`contextWindow`) or snake_case (`context_window`). The key must
contain a `/` — a bare model id is ignored because the registry is keyed by
`provider/model`. Documented in `docs/providers.md`
(**Overriding model metadata**) and cross-referenced from
`docs/configuration.md`.

## Verification

- New unit tests, all passing:
  - `crates/api` — `model_registry::tests::override_*` (4): patch existing
    catalog entry, partial-patch leaves other fields intact, synthesize entry
    for unknown alias, malformed/empty keys ignored.
  - `crates/core` — `config::tests::effective_config_merges_top_level_model_overrides`
    and `config::tests::model_override_accepts_camel_and_snake_case`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -41,6 +41,7 @@ values. Keys absent from the project file fall back to the global value.
   "provider": "anthropic",
   "config": { ... },
   "providers": { ... },
+  "modelOverrides": { ... },
   "projects": { ... },
   "commands": { ... },
   "formatter": { ... },
@@ -54,7 +55,9 @@ values. Keys absent from the project file fall back to the global value.
 ```
 
 Most day-to-day options live inside the `config` object. Provider credentials
-live in the `providers` map.
+live in the `providers` map. Corrected model metadata for self-hosted or
+unknown models lives in the `modelOverrides` map — see
+[Model metadata overrides](providers.md#overriding-model-metadata).
 
 ---
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -629,6 +629,16 @@ matches. They are defined in the `formatter` map:
     }
   },
 
+  // Correct metadata for self-hosted / unknown models (keyed by provider/model).
+  // Overrides win over the models.dev catalog.
+  "modelOverrides": {
+    "custom-openai/my-local-llm": {
+      "contextWindow": 32768,
+      "maxOutputTokens": 4096,
+      "name": "My Local LLM"
+    }
+  },
+
   // Custom slash commands
   "commands": {
     "test": {
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -711,3 +711,51 @@ The above example allows only `gpt-4o` (whitelist minus blacklist).
 Claurst ships a bundled snapshot of models for Anthropic, OpenAI, and Google. At runtime it optionally refreshes from the public `https://models.dev/api.json` API (cached to `~/.claurst/models_cache.json`, refreshed at most every 5 minutes). Network failures are swallowed silently; the bundled snapshot is always sufficient for normal operation.
 
 When no model is explicitly set, Claurst scores available models by priority patterns to pick the best default. Well-known model prefixes (`claude-*`, `gpt-*`, `gemini-*`, etc.) are always routed to their canonical provider regardless of gateway entries in the remote cache.
+
+### Overriding model metadata
+
+Self-hosted endpoints (the `custom-openai` / Ollama / LM Studio / llama.cpp
+providers) and model aliases that models.dev does not know can end up with the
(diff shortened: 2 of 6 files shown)


Kuberwastaken (commented):
## Review

**Design** — the right architecture: `ModelRegistry` is the single funnel for every context-window consumer (`refresh_context_window_size`, auto-compact's `resolve_context_window`, picker projections), so layering overrides there fixes all surfaces at once. Retaining the overrides inside the registry and re-asserting after `load_cache()` / `refresh_from_models_dev()` closes the clobber-on-refresh hole most implementations miss.

- ✅ Precedence (user > models.dev > builtin) implemented as in-place patching of `Some` fields only; partial overrides leave catalog values intact (tested).
- ✅ Synthetic entries for unknown aliases: sensible defaults (`tool_calling: true` so agent tools aren't disabled for self-hosted models; text modalities).
- ✅ Config plumbing mirrors the existing `providers`/`provider_configs` conventions: top-level + nested `config` block (nested wins, tested), camelCase/snake_case aliases, project-over-global `merge_map`.
- ✅ Live-discovery picker path (`ctx_for`/`name_for`) covered too — self-hosted endpoints show corrected windows even when the entry never touches the catalog.
- ✅ Malformed keys warn+skip instead of panicking; empty overrides are no-ops (tested).
- ✅ Docs: configuration.md + providers.md with concrete llama.cpp/Ollama examples; commit hygiene is clean (6 scoped conventional commits).

**Nits (non-blocking):**
- A bare-id override key (`"my-model"` without provider) is skipped with a tracing warn — fine, but users won't see tracing by default; the docs do state the required form, good enough.
- `run_models_command` re-loads settings for overrides while other callers thread `config` — acceptable given that path builds its own registry anyway.

Merging once CI is green on all three platforms.

PR #319, opened 2026-07-16
fix: honest cache accounting for OpenAI-compatible providers + local-models docs

### LINKED ISSUE #314: caching is not working for local models

Kuberwastaken:
## Problem

Reported in #314: running a local `llama-server` (OpenAI-compatible endpoint),
the cache accounting never reflects any activity, and there was no documentation
for driving Claurst against local models. The reporter also asked whether
`CLAURST_COORDINATOR_MODE=1` (suggested to them by an AI) is a real setting.

Two concrete issues under the hood:

1. The OpenAI-compatible provider hardcoded `cache_read_input_tokens` /
   `cache_creation_input_tokens` to `0` in `parse_usage`, so prompt-cache hits
   were thrown away even when the server reported them. OpenAI, LM Studio, vLLM,
   and llama.cpp (>= b4600 with prompt caching) all report reuse under
   `usage.prompt_tokens_details.cached_tokens`.
2. When a provider never reports cache info, `/usage` and `/extra-usage` showed a
   permanent `0`, which reads as "caching is broken" rather than "no data."

## Approach

**Cache parsing (`crates/api/src/providers/openai.rs`)**
- Parse `usage.prompt_tokens_details.cached_tokens` into `cache_read_input_tokens`.
- OpenAI's `prompt_tokens` is *inclusive* of cached tokens (unlike Anthropic,
  whose `input_tokens` excludes them), so the cached count is split back out of
  `input_tokens`, keeping the additive invariant
  `input_tokens + cache_read_input_tokens == prompt_tokens` that the cost tracker
  and context accounting rely on (no double counting). Clamped defensively so a
  malformed `cached > prompt` can't underflow.
- Both streaming and non-streaming paths go through `parse_usage`, so both are
  fixed. Added 5 unit tests.

**Honest counters (`crates/commands/src/usage.rs`)**
- `/usage` and `/extra-usage` now show `n/a` (and `n/a` cache-hit rate) when a
  session recorded zero cache activity, so "provider reports no cache info" is
  distinguishable from a real zero. Added 3 unit tests.

**Documentation (`docs/local-models.md`, new)**
- Connecting via llama.cpp and any OpenAI-compatible server (LM Studio, Ollama,
  vLLM).
- Recommended `llama-server` flags for agentic use: `--jinja` (required for
  tool-call templates), context sizing and the `--parallel` / `--ctx-size` split
  gotcha, prompt caching and `cached_tokens` reporting, the `--no-context-shift`
  tradeoff, and performance flags.
- Why tool calling needs a tool-aware chat template; conservative model guidance
  (no invented benchmarks).
- What the `0k/262k` status counter actually is (context window, not a cache
  counter) and how cache read/write surface in `/usage`.
- Wired into the docs sidebar/registry and cross-linked from the overview and
  providers pages.

**`CLAURST_COORDINATOR_MODE` — the honest answer (documented in the guide)**

Grepping the tree: the string exists **only** as a constant in
`crates/query/src/coordinator.rs`. `is_coordinator_mode()` (which reads it) has
**no callers in the live agent loop**, and `crates/query/src/runner/prompt.rs`
hardcodes `coordinator_mode: false`. So setting `CLAURST_COORDINATOR_MODE=1`
currently has no observable effect, and it is not needed — Claurst is agentic by
default. The docs say this plainly so nobody cargo-cults it.

## Verification

- `cargo clippy -p claurst-api -p claurst-commands --all-targets` — no warnings.
- `cargo test -p claurst-api -p claurst-commands --lib` — 107 + 66 pass,
  including the new `parse_usage_*` and `display_cache_count` tests.
- Docs loader confirmed to resolve `local-models` -> `local-models.md`.

## Notes

- The reporter's other symptom — pasted text arriving empty ("every prompt starts
  from the beginning, I have to repeat myself") — is a separate TUI paste bug
  being fixed in its own PR, not addressed here. That bug is the more likely
  reason the `0k/262k` context counter appeared stuck, since empty prompts never
  advance the context.

Closes #314

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -487,6 +487,10 @@
       <svg width="14" height="14" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="1.5"><circle cx="8" cy="8" r="6"/><path d="M8 2a5 5 0 010 12M2 8h12"/></svg>
       Providers
     </a>
+    <a class="sidebar-link" data-page="local-models" onclick="loadPage('local-models')">
+      <svg width="14" height="14" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="1.5"><rect x="2" y="3" width="12" height="7" rx="1"/><path d="M5 13h6M8 10v3"/></svg>
+      Local Models
+    </a>
     <a class="sidebar-link" data-page="mcp" onclick="loadPage('mcp')">
       <svg width="14" height="14" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="1.5"><rect x="2" y="6" width="5" height="4" rx="1"/><rect x="9" y="6" width="5" height="4" rx="1"/><path d="M7 8h2"/></svg>
       MCP Integration
@@ -551,6 +555,7 @@ const PAGES = [
   { id: 'tools',         title: 'Tools Reference' },
   { id: 'keybindings',   title: 'Keybindings' },
   { id: 'providers',     title: 'Providers' },
+  { id: 'local-models',  title: 'Local Models' },
   { id: 'mcp',           title: 'MCP Integration' },
   { id: 'plugins',       title: 'Plugins' },
   { id: 'agents',        title: 'Agents' },
--- a/docs/index.md
+++ b/docs/index.md
@@ -134,7 +134,9 @@ claurst --provider ollama --model llama3.2
 OPENAI_API_KEY=sk-... claurst --provider openai --model gpt-4o
 ```
 
-See [Providers](providers) for setup instructions for every supported provider.
+See [Providers](providers) for setup instructions for every supported provider,
+or the [Local Models](local-models) guide for running against llama.cpp,
+LM Studio, Ollama, or any OpenAI-compatible server on your own machine.
 
 ---
 
@@ -192,6 +194,7 @@ See [Slash Commands](commands) for the complete reference.
 - [Slash Commands](commands) — all 70+ commands
 - [Tools Reference](tools) — all 40+ tools and permission levels
 - [Providers](providers) — configuring each LLM provider
+- [Local Models](local-models) — llama.cpp, LM Studio, Ollama, and other OpenAI-compatible servers
 - [MCP Integration](mcp) — Model Context Protocol servers
 - [Plugins](plugins) — building and using plugins
 - [Agents](agents) — multi-agent orchestration
--- a/docs/local-models.md
+++ b/docs/local-models.md
@@ -1,0 +1,294 @@
+# Running Claurst against local models
+
+Claurst can drive any server that speaks the **OpenAI-compatible chat
+completions API** (`POST /v1/chat/completions`). That includes:
+
+- **llama.cpp** (`llama-server`) — provider id `llamacpp`
+- **LM Studio** — provider id `lmstudio`
+- **Ollama** (via its `/v1` OpenAI shim) — provider id `ollama`
+- **vLLM**, **text-generation-webui**, **LocalAI**, and anything else that
+  exposes `/v1/chat/completions` — use the generic `openai` provider with a
+  custom base URL
+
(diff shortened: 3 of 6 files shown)


Kuberwastaken (commented):
## Review

**The accounting fix is the correct one.** OpenAI-compatible `prompt_tokens` is *inclusive* of `prompt_tokens_details.cached_tokens`, while claurst's internal convention (from Anthropic) is additive — the PR splits the cached count back out (`input_tokens = prompt - cached`), preserving `input + cache_read == prompt_tokens` so the cost tracker and context accounting never double-count. Defensive `.min(prompt_tokens)` clamp prevents underflow on malformed servers. Five focused unit tests cover absent usage, absent details, zero, normal split, and the clamp.

- ✅ `/usage` & `/extra-usage` now distinguish **n/a** (provider never reported cache data) from a genuine 0 — this was the reporter's actual confusion in #314.
- ✅ Docs: docs/local-models.md is thorough and honest — the `CLAURST_COORDINATOR_MODE` answer is grep-verified against the source (constant exists, `is_coordinator_mode()` has no callers outside its own tests, system-prompt builder hardcodes false), the `--no-context-shift` cache-prefix tradeoff is explained correctly, and there's a practical curl to verify a server reports cache hits. No invented benchmarks.
- ✅ Scoped commits (api fix / display fix / docs), CI green on all 3 platforms.
- ✅ Correctly leaves the paste half of #314 to #316 and links the story together.

Non-blocking: the reporter's issue also asked about repeated re-thinking on every prompt — the docs' prefix-caching + `--cache-prompt` guidance addresses the practical cause.

PR #320, opened 2026-07-16
fix(api): echo Gemini thoughtSignature on tool calls (#311)

### LINKED ISSUE #311: Google Gemini 3.1 Pro Preview fails on Read tool with "Function call is missing a thought_signature"
## Environment

- OS: Windows 11
- Claurst Version: v0.1.7
- Provider: Google
- Model: Gemini 3.1 Pro Preview
- Billing: Enabled

## Description

Normal chat requests work correctly.

Example:

```
Explain what JWT authentication is.
```

returns a valid response.

However, any request that requires Claurst to read project files fails.

For example:

```
Read GEMINI.md
```

or

```
Read PROJECT_SPEC.md
```

The API returns:

```
API error: [google] Error 400:
Function call is missing a thought_signature in functionCall parts.
This is required for tools to work correctly...
Additional data:
function_call `default_api:Read`
position 2
```

## Steps to Reproduce

1. Connect Google provider.
2. Select Gemini 3.1 Pro Preview.
3. Open a project folder.
4. Ask Claurst to read a local file.
5. Observe the API error.

## Expected Behavior

Claurst should read the file and answer the prompt.

## Actual Behavior

Google returns HTTP 400 complaining about a missing `thought_signature` during the `Read` tool call.

## Additional Information

- Simple chat works.
- Tool/function calls appear to fail.
- Other errors previously observed:
  - Rate limited (before enabling billing)
  - Model not found (using older Gemini model IDs)

Image

Kuberwastaken:
## Problem

Google Gemini 3.x "thinking" models (e.g. `gemini-3.1-pro-preview`) reject **any**
tool call on the next turn with HTTP 400:

> Function call is missing a thought_signature in functionCall parts. This is
> required for tools to work correctly.

Plain chat works; the failure only appears once a tool (Read, etc.) is invoked.

Root cause: these models attach an opaque `thoughtSignature` to each `functionCall`
part in their response. Per Google's rules the client must echo that exact signature
back on the corresponding `functionCall` part when it replays conversation history.
Claurst dropped the field, so every follow-up turn was rejected.

## Approach

- Add a provider-agnostic `thought_signature: Option<String>` passthrough field to
  the internal `ContentBlock::ToolUse` type (`crates/core`). It is
  `#[serde(default, skip_serializing_if = "Option::is_none")]`, so other providers'
  wire formats and persisted sessions are byte-identical to before, and the field
  survives session save/load.
- Capture the camelCase `thoughtSignature` in the Google provider from **both** the
  non-streaming `generateContent` response and the streaming SSE parser (shared
  `thought_signature_from_part` helper).
- Carry it on the `ToolUse` block through stream assembly: the query-loop tool-call
  accumulator and the `StreamBlockAccumulator`.
- Re-emit it as a sibling of the `functionCall` part when serializing history back to
  Gemini `contents`. Absent signature keeps the old wire shape (no key emitted); no
  signature is ever fabricated for calls that never carried one.
- Mechanical fan-out of the new field to every other `ToolUse` construction site
  (all providers pass `None`).

## Verification

- No Gemini API key is available in this environment, so the fix was not exercised
  against the live API. Coverage relies on round-trip unit tests in
  `crates/api/src/providers/google.rs`:
  - `thought_signature_from_part_captures_streaming_signature` — streaming extraction
    (and absent → `None`).
  - `parse_response_body_captures_thought_signature` /
    `parse_response_body_omits_signature_when_absent` — non-streaming capture.
  - `build_request_body_round_trips_thought_signature_onto_function_call` /
    `build_request_body_without_signature_omits_thought_signature_key` — request
    serialization re-attaches the signature as a `functionCall` sibling, and omits the
    key entirely when absent.
  - `thought_signature_survives_content_block_serde_round_trip` — session save/load.
  - `provider_types` stream-accumulator test asserts the signature survives streaming
    assembly.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo test --workspace` — all pass (117 in `claurst-api`, full suite green).

Closes #311

### DIFF
--- a/src-rust/crates/api/src/lib.rs
+++ b/src-rust/crates/api/src/lib.rs
@@ -1488,7 +1488,7 @@ impl StreamAccumulator {
                         PartialBlock::ToolUse { id, name, json_buf } => {
                             let input = serde_json::from_str(&json_buf)
                                 .unwrap_or(Value::Object(Default::default()));
-                            ContentBlock::ToolUse { id, name, input }
+                            ContentBlock::ToolUse { id, name, input, thought_signature: None }
                         }
                         PartialBlock::Thinking {
                             thinking_buf,
--- a/src-rust/crates/api/src/protocol/openai_chat.rs
+++ b/src-rust/crates/api/src/protocol/openai_chat.rs
@@ -225,6 +225,7 @@ impl OpenAiChatDecoder {
                             id: tc_id.to_string(),
                             name,
                             input: json!({}),
+                            thought_signature: None,
                         },
                     });
                 }
--- a/src-rust/crates/api/src/provider_types.rs
+++ b/src-rust/crates/api/src/provider_types.rs
@@ -204,6 +204,9 @@ enum PartialBlock {
         id: String,
         name: String,
         json_buf: String,
+        /// Opaque provider signature (e.g. Gemini's `thoughtSignature`) captured
+        /// on `ContentBlockStart`; echoed back on the finalized block (#311).
+        thought_signature: Option<String>,
     },
     /// Any block that arrives fully-formed on `ContentBlockStart` and needs no
     /// delta accumulation (e.g. `RedactedThinking`, images).
@@ -225,10 +228,11 @@ impl PartialBlock {
                 thinking_buf: thinking,
                 signature_buf: signature,
             },
-            ContentBlock::ToolUse { id, name, .. } => PartialBlock::ToolUse {
+            ContentBlock::ToolUse { id, name, thought_signature, .. } => PartialBlock::ToolUse {
                 id,
                 name,
                 json_buf: String::new(),
+                thought_signature,
             },
             other => PartialBlock::Passthrough(other),
         }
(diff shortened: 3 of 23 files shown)


Kuberwastaken (commented):
## Review

**Correct diagnosis and the right mechanism.** Gemini 3.x thinking models attach an opaque `thoughtSignature` to `functionCall` response parts; it must be echoed back verbatim on the next turn or the API 400s exactly as reported. This PR captures it (streaming + non-streaming, via a shared `thought_signature_from_part` helper since both parts have identical shape) and re-attaches it as a sibling of `functionCall` — the correct position in the REST shape — when serializing history back to `contents`.

- ✅ Provider-agnostic passthrough: `thought_signature: Option<String>` on `ContentBlock::ToolUse` with `#[serde(default, skip_serializing_if)]` — old persisted sessions load unchanged, other providers' wire formats stay byte-identical, and the field survives session save/load (serde round-trip tested).
- ✅ Stream assembly threads it through `StreamBlockAccumulator`/`PartialBlock`; the lib.rs Anthropic-event accumulator is correctly left at `None` (thoughtSignature doesn't exist on that protocol).
- ✅ No fabricated signatures; absent signature emits the exact old wire shape (tested).
- ✅ Compact/session-memory clone sites preserve the signature, so history rewrites don't strip it.
- ✅ Test matrix: streaming capture, non-streaming capture, request round-trip, absent-key omission, persistence round-trip; `cargo test --workspace` fully green and `clippy --workspace --all-targets -D warnings` clean.
- ✅ Extra credit: caught and reverted CRLF line-ending noise a prior editing pass had injected into `api/lib.rs` (mixed-endings file), shrinking that diff to the single semantic line.

**Honest caveat (from the PR itself):** not exercised against the live Gemini API (no key in the build environment) — coverage is via the round-trip unit tests. Worth a quick manual confirmation by anyone with a Gemini key on the issue thread after merge.

Merging on green CI.

PR #325, opened 2026-07-21
fix: claurst upgrade fails with Text file busy (ETXTBSY)

sammyyakk:
## Summary

`claurst upgrade` reliably fails on Linux/macOS with:

```
Error: failed to copy new binary into /home/user/.claurst/bin/claurst

Caused by:
    Text file busy (os error 26)
```

Root cause is in `swap_binary()` (`src-rust/crates/cli/src/upgrade.rs`). The unix path uses `std::fs::copy(new, current)`, which opens the existing file with `O_TRUNC` and writes into its current inode. Since `current` is the running process's own executable, that inode is mapped as the process's text segment, and the kernel rejects the write with `ETXTBSY`.

The comment above the old code claimed the kernel allows this because "unlink-and-replace just frees the directory entry" — that description matches `rename()` semantics, not `copy()`. The code and the comment didn't match, which is the actual bug.

## Fix

Stage the new binary next to the current one (`claurst.new`), `chmod 755` it, then `rename()` it into place. `rename()` repoints the directory entry to a new inode instead of writing into the old one, so it succeeds even while the old binary is still running — no ETXTBSY.

## Test plan

- [x] Added a regression test (`upgrade::tests::swap_binary_replaces_a_running_executable`) that copies a real ELF binary, spawns it so its text segment is mapped, and calls `swap_binary` on it while it's running. The test also asserts that a direct `fs::copy` overwrite fails first, confirming the fixture reproduces the real ETXTBSY condition.
- [x] `cargo test -p claurst` — all 13 tests pass.
- [x] `cargo build -p claurst --release` — builds clean.

### DIFF
--- a/src-rust/crates/cli/src/upgrade.rs
+++ b/src-rust/crates/cli/src/upgrade.rs
@@ -293,6 +293,68 @@ fn walkdir_shallow(root: &Path, max_depth: usize) -> Vec<PathBuf> {
 // Atomic binary swap
 // ---------------------------------------------------------------------------
 
+#[cfg(unix)]
+struct StagedBinary {
+    path: PathBuf,
+}
+
+#[cfg(unix)]
+impl StagedBinary {
+    fn create_next_to(current: &Path) -> Result<(Self, std::fs::File)> {
+        use std::ffi::OsString;
+        use std::fs::OpenOptions;
+        use std::io::ErrorKind;
+        use std::sync::atomic::{AtomicU64, Ordering};
+
+        static NEXT_STAGE_ID: AtomicU64 = AtomicU64::new(0);
+        const MAX_ATTEMPTS: usize = 128;
+
+        let parent = current
+            .parent()
+            .ok_or_else(|| anyhow!("installed binary path has no parent: {}", current.display()))?;
+        let file_name = current.file_name().ok_or_else(|| {
+            anyhow!(
+                "installed binary path has no file name: {}",
+                current.display()
+            )
+        })?;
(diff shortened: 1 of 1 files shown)


Kuberwastaken (changes requested):
The ETXTBSY diagnosis and same-filesystem rename approach are correct, but this needs changes before merge:

1. The regression test hardcodes `/usr/bin/sleep` (`upgrade.rs:377-378`). macOS CI fails because that path does not exist on the runner. Use a portable fixture or resolve a suitable executable per platform.
2. `current.with_extension("new")` creates one predictable shared staging path (`upgrade.rs:326`). Concurrent upgrades can truncate/write the same inode, and a pre-existing `claurst.new` is followed or overwritten. Create a unique same-directory file with exclusive creation, write and sync it, then rename.
3. The `chmod` result is discarded (`upgrade.rs:329-332`) before the staged file is installed. A chmod failure can replace Claurst with a non-executable file. Set Unix permissions through `PermissionsExt`, propagate failures before rename, and assert replacement content plus executable mode in the test.

Ubuntu CI passes; macOS currently fails on item 1. Please address these and rerun the full matrix.

sammyyakk:
All three review points addressed in 5bd767f — unique exclusive-create staging file with Drop cleanup, propagated `set_permissions` via `PermissionsExt` before rename, and a portable test fixture (copies its own test binary instead of `/usr/bin/sleep`). CI green on ubuntu/macos/windows now. Looks good to merge.

PR #327, opened 2026-07-24
feat(tui): add configurable external status line

Nick768:
## Summary

Adds a configurable external status line feature to Claurst, analogous to Claude Code's `statusLine`. The status line runs a user-defined command periodically and displays its stdout in a dedicated row between the prompt input and the footer.

### Core (settings.json)
- New `statusLine` config block under `config`: `command`, `pollIntervalMs`, `padding`
- `#[serde(default)]` on `Config` to allow partial deserialization (minimal settings files)
- Backward compatible: `CLAUDE_STATUS_COMMAND` env var continues to work

### CLI/TUI
- Background task polls the command at the configured interval, passes session JSON on stdin
- Session data includes model, context window usage, cost, workspace info
- Rendered as a separate layout row (not inline in the footer)
- ANSI escape sequences stripped for clean display
- `/cc-statusline` slash command shows current external status line config
- Padding support via `statusLine.padding`

### Dev UX
- `shell.nix` for NixOS builds (includes all native dependencies)

## Testing
- Unit tests for `StatusLineConfig` deserialization
- Test for `/cc-statusline` command
- All existing tests pass (657 TUI, 500 core)

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,15 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+pkgs.mkShell {
+  buildInputs = with pkgs; [
+    cargo rustc rustfmt clippy
+    gcc cmake pkg-config gnumake perl
+    openssl.dev alsa-lib.dev
+    clang libclang.lib libclang.dev
+    python3
+    libxkbcommon wayland
+  ];
+
+  LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
+  PKG_CONFIG_PATH = "${pkgs.alsa-lib.dev}/lib/pkgconfig";
+}
--- a/src-rust/crates/cli/src/main.rs
+++ b/src-rust/crates/cli/src/main.rs
@@ -1977,32 +1977,49 @@ async fn run_interactive(
         }
     }
 
-    // CLAUDE_STATUS_COMMAND: optional external command whose stdout replaces the
-    // left-side status bar text. Polled every 500ms (debounced) in the main loop.
-    // The command is run in a background task; results flow through a channel.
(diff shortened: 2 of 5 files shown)


Kuberwastaken (changes requested):
This cannot merge safely in its current form:

1. Project-level `.claurst/settings.json` is loaded during hierarchical config resolution and its `statusLine` overrides the global value (`core/src/lib.rs:1862-1900,1978`). The command is then executed automatically through `sh -c`/`cmd /C` (`cli/src/main.rs:1992-2012`). Opening an untrusted repository can therefore execute arbitrary code without approval. Make this setting global-only, or add an explicit project-command trust gate and regression test.
2. `pollIntervalMs` is documented with a 1000 ms minimum but is not validated; `0` creates a tight subprocess loop. The task and child have no cancellation ownership or timeout, and output capture is unbounded. Clamp/validate the interval, stop the task with the session, kill timed-out children, and bound captured output.
3. Child stderr is inherited while ratatui owns the terminal, so command failures can corrupt the TUI. Pipe or null stderr and surface errors through application state.
4. User-controlled `u16` padding uses unchecked `area.x + padding` and `padding * 2` (`tui/src/render.rs:2858-2863`), which can overflow or create an invalid rectangle. Clamp to the available width and add narrow-terminal/large-padding tests.
5. The advertised session payload sets an empty `session_id`, reports all tokens as input, and always reports zero output tokens even though the real session ID and split counters are available.
6. The ANSI stripping only handles SGR sequences ending in `m`; OSC, other CSI sequences, controls, and multiline output are not safely normalized for a one-row layout.

Please also remove the unrelated `shell.nix` from this feature PR. Once the execution trust boundary and lifecycle are corrected, the feature can be reviewed on its merits.

PR #331, opened 2026-07-29
feat(core): add named workspace roots

tykkidream:
## Motivation

`--add-dir` already allowed Claurst to access extra directories, but those directories were mostly anonymous permission roots. In multi-repository workflows, the model had to rely on absolute paths or user reminders to search and edit outside the main working directory.

This change makes the main directory and additional directories visible as named workspace roots, so users and the model can refer to them consistently with `&main` and `&root-name/path`.

`&` is used instead of `@` because `@` is already used by prompt file injection.

## Summary

- Introduce named workspace roots, with `&main` for the primary working directory and stable sanitized names for additional directories from `--add-dir`, `additional_dirs`, and `workspace_paths`.
- Add `&root-name/path` resolution through the shared tool path resolver, so path-based tools can target additional roots without requiring absolute paths.
- Expose the workspace root list to the model prompt, including guidance for searching across roots with parallel `Glob` / `Grep` calls.
- Update the TUI welcome screen to show named roots instead of anonymous `+dir` entries.
- Document workspace-root paths in the advanced guide, configuration reference, and tools reference.

## Test plan

- `cargo check --workspace`
- `cargo test --package claurst-core workspace`
- `cargo test --package claurst-core system_prompt`

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -637,6 +637,10 @@ Grants Claurst read access to an additional directory outside the working direct
 
 Multiple `--add-dir` flags can be combined.
 
+The working directory and additional directories are exposed to the model as named workspace roots. The primary working directory is always `&main`; each additional directory gets a stable root name derived from its directory name, lowercased and sanitized for use in prompts. For example, `_ai-engine` becomes `&_ai-engine`, and `My Project (API)` becomes `&my-project-api`.
+
+Path-based tools accept absolute paths, paths relative to `&main`, and workspace-root paths such as `&main/src/lib.rs` or `&_ai-engine/prd/spec.md`. For `Glob` and `Grep`, omitting `path` searches only `&main`; to search all workspace roots, the model calls the tool once per root, which can run in parallel.
+
 ### Environment variables in config
 
 Environment variables can be set in `.claude/settings.json` under an `env` key. These are injected into tool executions:
@@ -680,7 +684,7 @@ The `LspTool` provides code intelligence by communicating with a Language Server
 ```
 
 - `action` — required, one of the operations above
-- `file` — required, absolute or working-directory-relative path
+- `file` — required, absolute, working-directory-relative, or `&root-name/relative` path
 - `line` — 1-based line number (required for `hover`, `definition`, `references`)
 - `column` — 1-based column number (required for `hover`, `definition`, `references`)
 
@@ -707,4 +711,4 @@ LSP servers must be configured in `settings.json`:
 }
 ```
 
-If no LSP server is configured for the file's language, the tool returns an informative error. Path resolution is relative to the current working directory.
+If no LSP server is configured for the file's language, the tool returns an informative error. Relative paths resolve against `&main`; `&root-name/relative` paths resolve against the named workspace root.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -120,7 +120,7 @@ prefixed with their server name (`myserver_toolname`).
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| `additional_dirs` | array of strings | [] | Additional filesystem paths Claurst is allowed to read and write. Equivalent to passing `--add-dir` on the command line. |
+| `additional_dirs` | array of strings | [] | Additional filesystem paths Claurst is allowed to read and write. Equivalent to passing `--add-dir` on the command line. These directories are also exposed as named workspace roots alongside `&main`, so path-based tools can target them with `&root-name/relative` paths. |
 
 ### MCP servers
 
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -39,6 +39,12 @@ Every tool in Claurst implements a common `Tool` interface. This interface defin
 
 Tools are loaded eagerly at session start. The model receives tool descriptions and schemas and selects tools to call. Each tool call goes through permission resolution before `call()` is invoked.
 
+### Workspace-root paths
+
+Path-based tools accept absolute paths, paths relative to the main workspace root, and named workspace-root paths. The primary working directory is always available as `&main`; additional directories from `--add-dir` or `additional_dirs` are assigned sanitized names such as `&_ai-engine` or `&my-project-api`.
+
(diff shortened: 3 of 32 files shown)


Kuberwastaken (changes requested):
Thanks — this fills a real gap (main never tells the model about `--add-dir` roots), the name-generation/parsing tests are good, and the permission story is fine (`&root/` resolution stays inside already-allowed roots). Not mergeable as-is though:

1. clippy `-D warnings` fails in claurst-core: `workspace.rs:58` `manual_pattern_char_comparison` (`.find(['/', '\\'])`), `:199`/`:227` clone-to-slice in tests. Expect more in tools/tui once core passes (`&ctx.main_root()` patterns).
2. Replacing `ToolContext.working_dir` with a `BTreeMap` + `main_root()` doing `.get("main").expect(..)` puts a map lookup and a panic path on every tool's hot path, and three sites in `cli/main.rs` (1845, 2474, 2562) now have to remember to re-insert `"main"` when cwd changes. That's why this touches 32 files. Keep `working_dir: PathBuf` and add `extra_roots` alongside it (or a `WorkspaceRoots { main, extras }` struct) — no `expect`, no manual re-insert, ~10-file diff.
3. `generate_root_names` `assert!(primary.is_absolute())` (`workspace.rs:84`) is reachable from the ACP prompt handler with a client-supplied relative cwd → panic. Return `Result` or normalise.
4. `system_prompt.rs:324` emits the whole workspace-roots paragraph whenever roots are non-empty, and the CLI always inserts `main`, so every session pays for multi-root instructions that don't apply. Gate on non-main roots.
5. Unused surface: `WorkspaceRoot`, `ToolContext::root()`, `format_workspace_path`, `find_root_for_path`, and the `ResolvedWorkspacePath` fields are never read. Delete, and avoid the `String`/`PathBuf` allocations in `resolve_workspace_path` per call.
6. Add one end-to-end test in crates/tools: `resolve_workspace_path("&docs/x")` → resolved path → permission check. Current tests cover only the pure helpers.

Happy to review again after that.

Kuberwastaken (changes requested):
Thanks for this — it's a thoughtful PR and the motivation is real. Multi-repo sessions today do force the model into absolute paths, and `&main` / `&root/path` is a nice affordance. The docs updates and the `workspace.rs` test suite are appreciated. A few things I'd want addressed before merging.

**First, heads up on CI:** there are no check runs on this branch. The one workflow run recorded for it failed at startup with zero jobs (fork-PR Actions gating), so nothing here has been compiled, tested, or linted on any platform — and the run is too old to retry from my side. I'll approve CI on your next push; I expect clippy to be red, since `ci.yml` runs `cargo clippy --workspace --all-targets -- -D warnings`:
- ~9 `needless_borrow` sites where `&ctx.working_dir` became `&ctx.main_root()` — that's now `&&PathBuf`. Drop the `&`: `cli/src/main.rs` (`parse_at_refs`), `query/src/agent_tool.rs` (`find_git_root`), `query/src/lib.rs` ×4, `tools/src/pty_bash.rs`, `tools/src/worktree.rs` ×2.
- `field_reassign_with_default` in the new `render.rs` test — use a struct literal with `..Default::default()`.

**Security — the one that needs a design change.** A project's own `.claurst/settings.json` can append to `workspace_paths` and `additional_dirs` (`core/src/lib.rs:1960-1961`); unlike `trust_project_mcp_servers` and `allowed_bash_prefixes`, those aren't protected from project override. That's a known gap (#389). Today it silently widens the read auto-allow set. With this PR, those attacker-supplied paths also get a friendly name *and* land in the system prompt alongside "When the user asks to find or search something without naming a specific root, search all workspace roots." A hostile repo shipping `workspace_paths: ["/home/you/.ssh"]` now gets `&ssh` advertised to the model with a standing instruction to search it. Could we expose only roots that came from the CLI `--add-dir` (or from an explicit `/add-dir`), and leave project-settings-derived paths unnamed until #389 lands a trust gate?

Related: `build_workspace_roots_section` interpolates paths into `<root name="{}">{}</root>` with no escaping. Directory names can contain `<`, `>`, `"` and newlines, so a crafted path can close the tag and inject prompt text. Please escape both fields.

**Please keep `working_dir`.** Replacing the field with a `BTreeMap` + `main_root()` is what makes this a 32-file diff, and it introduces two panics: `main_root()`'s `.expect(...)` on a `pub` map anyone can mutate, and the `assert!` in `generate_root_names`. In `acp/src/prompt.rs` the primary comes from client-supplied `session.cwd`, so a relative cwd panics the ACP server — we've been going the other way lately (#333, #334). Keeping `working_dir` as-is and adding `named_roots: BTreeMap<String, PathBuf>` alongside it gets you the same feature, no panics, a fraction of the diff, and far fewer conflicts with #386 (which also touches `QueryConfig`, `cli/src/main.rs`, `query/src/lib.rs` and `render.rs`).

**Smaller things:**
- `resolve_workspace_path` doesn't normalize `..`, and if a `workspace_paths` entry is relative it returns a *relative* path — `resolve_path` used to always return an absolute one. Please normalize and assert absoluteness on the way out.
- `/add-dir` mid-session pushes to `config.workspace_paths` but never refreshes `tool_ctx.workspace_roots`, so access widens while the name never appears. Either rebuild the map or route both through one place.
- `find_root_for_path` returns the first alphabetical match, so with nested roots a file under the inner root reports the outer one — pick the longest matching root. It and `format_workspace_path` are also currently unused; either wire them up or drop them.
- `lsp_tool.rs` resolves `&otherroot/file` but still passes `ctx.main_root()` as the LSP workspace root for every action.
- The `<workspace_roots>` block is emitted whenever the map is non-empty, which is always (`main` is always there) — so single-root sessions now pay ~90 words of `&`-syntax guidance on every request. Worth gating on `len() > 1`, the way the TUI helper already does.
- Could you add tests for `ToolContext::resolve_workspace_path` itself? Specifically `&main/../../etc/passwd`, an unknown `&root` falling through to a literal filename, and a root whose path is relative. `workspace.rs` covers name generation well, but the resolver is the security-relevant half and is currently untested.

Also worth rebasing onto `main` — you're 6 commits behind. Happy to take this once the above are in.

Issue #186, opened 2026-06-20
[FEATURE REQUEST] /plugin marketplace add _?_


lucahttp:
is this possible?
is there any workaround?

Kuberwastaken:
beyond scope atm and there's no real standards afaik, open to suggestions though :)

Issue #57, opened 2026-04-07
Discuss: Future of claurst


haingo87:
As an optimization enthusiast who is obsessed with low-level languages, I truly appreciate your work. However, I have concerns about the future of this project regarding the fast moving pace of AI landscape:

- Are you planning to adapt claurst to Claude Code future features which are useful (I suppose not all of them are)?
- Do you have a plan for how claurst will evolve? Such as a roadmap, list of ideas, or building a community of contributors, etc.
- If you haven't thought about the future of claurst (to answer 2 questions above), is this project just one of your hobbies?

Kuberwastaken:
Pinning this - will be opening up the discussion tab soon to get a better sense of what people want from their TUI but if you have any thoughts/ suggestions before that, I'll be happy to address them here.

haingo87:
Yep, staying opened and gathering ideas from people actually using claurst is always the best. I would like to give one suggestion which is providing guideline (kind of Wiki page) and community (Reddit, discord channels, etc). It's simply to help people get to know claurst better. I know you there is already command line help, and the program is purely a port from Claude Code (that means there are many things people have already known). But the key differences are mentioned too briefly (even on github page) to start with.

For example, I'm trying to setup skills and agents inside .claurst folder and claurst doesn't recognize them. Obviously, an Wiki page would help here, even if it contains only differences in functions, configs, etc. with the original Claude Code, no need for a full manual. I believe with the help of AI, documentation about the difference is an easy part for you, considering how you utilize it to porting the whole source code.

About building community, I think just open one on Discord or reddit, etc and give the links in this github. My idea is simply getting more people onboard with this tool before we can gather contributions from them.

Kuberwastaken:
@haingo87 I get where you're coming from, docs are being worked on! (there's an early version [here](https://claurst.kuber.studio/docs) but not sure about maintaining communities haha, it'll be hard for me to manage though I 100% know where you're coming from :)

AdamBajger:
@haingo87 there is actually a lot of things that are currently:
 - ignored or not supported at all
 - overriden by hardcoded anthropic defaults
 - not delegated or shared properly 

I am trying to address these for the openly available providers and along the way maybe fix some of the more general issues, but it takes time. I assume the issues you are describing are not issues related to missing docs, but just features which need some care before being viable. Just stay tuned, or maybe even add your own contribution. In the meantime, opening a well described issue would help as well.

haingo87:
[...]

PR #336, opened 2026-07-31
feat(vscode): add a VS Code extension (ACP client, no reverse-engineering involved)

sammyyakk:
## Summary

> **Merge order:** this PR depends on #337 (Config tool `provider`/`effort` support). #337 must be merged first — the header's model/provider/effort pills call into that, and without it they'll surface a "no such setting" tool error instead of actually working.

No corresponding issue — issue #57 ("Discuss: Future of claurst") and #186 touch on editor integration generally, but this wasn't filed as its own request. Opening it as a concrete proposal.

Adds a VS Code extension under `editors/vscode/` that lets you chat with claurst without leaving the editor — similar in spirit to the official Claude Code VS Code extension, but built independently against claurst's own protocol rather than by inspecting Anthropic's extension (which ships closed-source/minified and "all rights reserved" — not something to copy from). Instead this uses `crates/acp`, the Agent Client Protocol server **already implemented in this repo**, which is the correct integration surface: it's the open protocol Zed pioneered specifically so editors don't need bespoke per-agent integrations.

## How it works

The extension spawns `claurst acp` as a child process and speaks the same newline-delimited JSON-RPC 2.0 wire format implemented in `src-rust/crates/acp/src/connection.rs`:

1. `initialize` / `session/new` handshake with the workspace folder as `cwd`.
2. `session/prompt` sends the user's message.
3. `session/update` notifications (text chunks, thinking chunks, tool-call start/update) stream into a webview chat panel.
4. Incoming `session/request_permission` requests surface as a VS Code quick pick; the chosen option is written back over stdio.

Every field name and enum tag used in `acpClient.ts` (`sessionUpdate`, `outcome`, `toolCallId`, etc.) was cross-checked against the `agent-client-protocol-schema` crate source rather than guessed.

## Commands

- **Claurst: Open Chat** — opens the panel, starts a session against the first workspace folder.
- **Claurst: New Session** — discards state, starts fresh.
- **Claurst: Stop Current Turn** — sends `session/cancel`.

Config: `claurst.executablePath` (default `"claurst"`, resolved from `PATH`).

## Update: UI overhaul + working folder + model/effort/provider controls

- Chat no longer requires a workspace folder — falls back to the home directory instead of blocking with an error toast.
- Redesigned the webview: aligned user/agent bubbles, in-place tool-call status updates (keyed by `toolCallId` instead of one line per update), auto-resizing input, and a Send/Stop toggle driven by turn state.
- Added header pills for **model**, **provider**, and **effort**. Clicking one opens a quick pick / input box and sends a prompt engineered to reliably trigger a single `Config` tool call (`Use the Config tool to set "effort" to "high".`) rather than relying on a conversational reply. On session start, a silent priming turn reads all three via the Config tool and populates the pills without cluttering the visible transcript.
- This is why #337 is a hard dependency: the `Config` tool didn't support `provider` or `effort` at all before that PR.

## Test plan

- [x] `npm install && npm run compile` in `editors/vscode/` — clean TypeScript build, no errors.
- [x] Smoke-tested the compiled `AcpClient` directly against a real running `claurst` binary (not just compiled — actually exercised): `initialize` → `session/new` → `session/prompt` round-trips correctly, streamed text chunks arrive, and a tool-use turn (`Bash`) correctly triggers a `tool_call` update, a `session/request_permission` round-trip (approved via the same code path the quick pick uses), a `tool_call_update` to `completed`, and the tool's real output streaming back.
- [ ] Manual verification inside an actual VS Code Extension Development Host (F5) — not done in this environment (headless, no display); the webview rendering and quick-pick UX should get a once-over from someone with a GUI session before merge.

## Scope / follow-ups

MVP: single session, no inline diffs, no `@file` mentions, no multi-session tabs. Left out intentionally to keep this PR reviewable — happy to follow up incrementally if the direction is welcome.

### DIFF
--- a/.github/workflows/vscode-extension-ci.yml
+++ b/.github/workflows/vscode-extension-ci.yml
@@ -1,0 +1,42 @@
+name: VS Code Extension CI
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'editors/vscode/**'
+      - '.github/workflows/vscode-extension-ci.yml'
+  pull_request:
+    branches: [main]
+    paths:
+      - 'editors/vscode/**'
+      - '.github/workflows/vscode-extension-ci.yml'
+
+concurrency:
+  group: vscode-extension-ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: editors/vscode
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+          cache-dependency-path: editors/vscode/package-lock.json
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Compile
+        run: npm run compile
+
+      - name: Unit tests
+        run: npm test
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,8 @@ refs/
 .vscode/
 .claurst/
 
+# ...except the VS Code extension's own launch/build config, which is
+# required (not personal editor prefs) for F5 debugging to work.
+!editors/vscode/.vscode/
+!editors/vscode/.vscode/*.json
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,11 @@
 # Rust
 src-rust/target/
 
+# VS Code extension
+editors/vscode/node_modules/
+editors/vscode/out/
+editors/vscode/*.vsix
+
 # OS
 .DS_Store
 
--- a/editors/vscode/.vscode/launch.json
+++ b/editors/vscode/.vscode/launch.json
@@ -1,0 +1,14 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Run Claurst Extension",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
+      "outFiles": ["${workspaceFolder}/out/**/*.js"],
+      "preLaunchTask": "npm: compile"
+    }
+  ]
+}
--- a/editors/vscode/.vscode/tasks.json
+++ b/editors/vscode/.vscode/tasks.json
@@ -1,0 +1,11 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "type": "npm",
+      "script": "compile",
+      "group": { "kind": "build", "isDefault": true },
+      "problemMatcher": ["$tsc"]
+    }
+  ]
+}
--- a/editors/vscode/README.md
+++ b/editors/vscode/README.md
@@ -1,0 +1,47 @@
+# Claurst for VS Code
+
+Chat with [Claurst](https://github.com/Kuberwastaken/claurst) without leaving VS Code.
+
+This extension does not reimplement the agent — it spawns `claurst acp` (the
+Agent Client Protocol server already built into the `claurst` CLI, see
+`src-rust/crates/acp`) as a child process and speaks newline-delimited
+JSON-RPC 2.0 to it over stdio. All prompting, tool execution, and permission
+logic live in `claurst` itself; this extension is a thin ACP client plus a
+webview for rendering the conversation.
+
+## Requirements
+
+- The `claurst` CLI installed and on `PATH` (or configure
+  `claurst.executablePath` to an absolute path).
+- A workspace folder open (the session's `cwd` is the first workspace folder).
+
+## Commands
+
+- **Claurst: Open Chat** — opens the chat panel and starts a session.
+- **Claurst: New Session** — discards the current session and starts fresh.
+- **Claurst: Stop Current Turn** — sends `session/cancel` for the in-flight turn.
+
+## Development
+
+```bash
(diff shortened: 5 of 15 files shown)


sammyyakk:
UI is functional but basic (vanilla HTML/CSS/JS, no design pass). If anyone wants to take the visual design further — better bubble/tool-call styling, icons, animations, general polish — very welcome. I've done my part on the protocol/functionality side; UI contributions appreciated.

Kuberwastaken (changes requested):
Thanks for building this against the ACP server rather than reverse-engineering anything; the acpClient.ts routing matches crates/acp. Can't merge it in this form though:

1. The model/provider/effort pills don't work against main: `captureStatusFromToolResult` requires `title === 'Config'`, but our `tool_call_update` never carries a title (prompt.rs builds status+content only and the schema skips None), and the initial `tool_call` has no content. So the pills never populate and the silent priming turn at session start is a paid LLM call for nothing. Same root cause makes `upsertToolCall` replace every tool line with "(tool call)" on completion.
2. Even when parsed, "set X via a prompt that asks the model to call the Config tool" only writes settings.json; the ACP runtime builds its QueryConfig once per process (runtime.rs:90), so the running session keeps the old model/effort. That needs a real mechanism (restart the child after a set, or a protocol-level option), not prompt injection. Drop the pills + priming turn from this PR and keep it to the chat/permission MVP.
3. On quick-pick dismissal default to the reject/cancel option (or `outcome: "cancelled"`), not `options[0]` — that's currently "allow once" for Bash/Edit approvals.
4. Add a CI job (`npm ci && npm run compile`, path-filtered to editors/vscode/) and at least a unit test for the line router; and confirm it has been run once in a real Extension Development Host.

Separately I need to decide whether an npm project belongs in this repo at all vs a `claurst-vscode` repo linked from the README; will follow up once the above is addressed.

sammyyakk:
All four addressed:

1. Confirmed the bug — \`tool_call_update\` never carries a title (\`prompt.rs\` only sets status/content on completion), so \`captureStatusFromToolResult\`'s \`title === 'Config'\` check silently no-opped and \`upsertToolCall\` blanked the line to "(tool call)" on completion. Fixed the display bug (element now remembers its title from the initial \`tool_call\` event and keeps it across updates with no title) — but per point 2 below, dropped the pills/priming feature that depended on this entirely rather than just patching it.
2. Dropped the model/effort/provider pills and the silent priming turn. Agreed a prompt-injected "set" only persists to settings.json while the running session's QueryConfig is already built — it wouldn't actually change live behavior, so it doesn't belong in this PR. Back to the chat/permission MVP.
3. Fixed — quick-pick dismissal (or no handler) now returns \`undefined\` from \`onRequestPermission\`, and \`AcpClient\` sends a real \`{outcome: {outcome: "cancelled"}}\` instead of guessing \`options[0]\`. Verified: a dismissed picker now results in the agent treating the tool call as denied, not allowed.
4. Added \`.github/workflows/vscode-extension-ci.yml\` (path-filtered to \`editors/vscode/\`, runs \`npm ci\`, \`npm run compile\`, \`npm test\`). Extracted the line-routing logic out of \`AcpClient\` into a pure \`acpProtocol.ts\` (parseLine/extractText, no child_process) specifically so it's unit-testable, and added 12 tests covering response/request/notification classification, malformed/blank/non-object input, and content extraction — using Node's built-in test runner, no new dependency. Ran \`npm ci && npm run compile && npm test\` clean. Also re-verified the full chat/permission flow against a real running \`claurst\` binary after all these changes (tool call → permission approval → completion → result text all still work).

No opinion from me on the repo-location question (this repo vs. a linked \`claurst-vscode\` repo) — happy to move it if that's the call.

Kuberwastaken:
Merged — this is nice work, and the ACP wiring holds up. I spot-checked every method and field against `agent-client-protocol-schema` 0.13 rather than the description: `initialize`, `session/new`, `session/prompt`, the `sessionUpdate` tags, the flattened `toolCallId`/`title`/`status` reads, and the `{outcome:{outcome:…}}` response shape are all correct, and `session/cancel` as a notification is right too. Still matches after #335. Vendoring it in `editors/vscode/` is the right home — the client and `crates/acp` should version together, and the path-filtered workflow plus the `acpProtocol` unit tests mean it won't rot silently.

Four things I'm fixing up in follow-up commits rather than another review round:

1. **Concurrent permission requests.** `spawn_drainer` (`crates/acp/src/permission.rs:161-171`) `tokio::spawn`s each `forward_pending` independently, so a parallel tool batch produces several `session/request_permission` calls at once. `showQuickPick` is a singleton in VS Code — opening the second dismisses the first, which returns `undefined`, which you (correctly) send as `Cancelled`, which the agent maps to `Deny`. Net effect: in a parallel batch, everything but the last tool gets silently denied. I'm adding a promise-chain queue so prompts show one at a time.
2. **README drift**: it still lists an open workspace folder as a requirement (`chatPanel.ts` falls back to `os.homedir()` now), and still says dismissing the picker grants the least-privileged option — your `6022e20` changed that to `Cancelled`, which is the better call; updating the doc to match. Also adding a line making clear this is build-from-source/F5 only for now (`publisher: "claurst"` isn't a registered marketplace identity yet, and there's no `.vscodeignore`).
3. CSP nonce → `crypto.randomUUID()`.
4. Surfacing `stopReason` from the `session/prompt` response so a refusal or turn-limit stop doesn't render like a clean finish.

Thanks for the contribution — and for doing it against the protocol instead of reverse-engineering anything.

PR #347, opened 2026-08-01
feat: default model option in model picker

alecuba16:
## Summary

Adds a `Default` option to model picker — clears model override, persists across reboots.

### Features
- Default entry at top of picker list
- Clears model override when selected
- Persists selection across restarts

### Files Changed
- `src-rust/crates/tui/src/model_picker.rs` — default option rendering
- `src-rust/crates/tui/src/app.rs` — default selection handler


alecuba16:

PR #351, opened 2026-08-01
fix: intercept bang commands in main loop + render output

alecuba16:
## Summary

Fixes bang commands not being intercepted (was dead code in `App::run`) and output not rendering when transcript empty.

### Features
- Intercept in `main.rs` submit flow (was dead code in `App::run`)
- Render output via `render_bang_display_lines()` when transcript empty
- `transcript_empty` check includes `display_messages.is_empty()`

### Files Changed
- `src-rust/crates/cli/src/main.rs` — bang intercept
- `src-rust/crates/tui/src/render.rs` — `render_bang_display_lines()`


alecuba16:

PR #354, opened 2026-08-01
fix: preselect correct row + match provider-prefixed model

alecuba16:
## Summary

`selected_idx` from grouped position, `is_current` matches provider-prefixed form. 5 tests.

### Files Changed
- `src-rust/crates/tui/src/model_picker.rs` — preselection logic + 5 tests


alecuba16:

PR #355, opened 2026-08-01
fix: picker group headers use registry provider names

alecuba16:
## Summary

`provider_names` HashMap from `model_registry.list_providers()`, render checks map first for group header names.

### Files Changed
- `src-rust/crates/tui/src/model_picker.rs` — `provider_names` HashMap + render check


alecuba16:

PR #356, opened 2026-08-01
fix: capture provider before close, skip default header

alecuba16:
## Summary

Captures `selected_provider_id` before close, skips `Other` header for default entry.

### Files Changed
- `src-rust/crates/tui/src/app.rs` — `selected_provider_id` capture before close


alecuba16:

PR #371, opened 2026-08-02
fix(tui): capture streaming text length at flush for TPS fallback

alecuba16:
## Summary

Fixes TPS (tokens per second) not displaying for providers that don't report `output_tokens` in the stream (many OpenAI-compatible providers).

## Root cause

`MessageStop` event arrives before `TurnComplete` and calls `flush_streamed_assistant_message()`, which does `std::mem::take(&mut self.streaming_text)` — emptying the buffer. `TurnComplete` then computed the token-count fallback via `self.streaming_text.len()` (now `0`), so TPS was never recorded.

## Fix

Added `last_flushed_stream_len: usize` to `App`, captured in `flush_streamed_assistant_message()` before `mem::take`. `TurnComplete` fallback now reads `last_flushed_stream_len` instead of the empty buffer.

## Test plan

- `cargo test --package claurst-tui -- test_turn_complete_tps` — verifies TPS records when provider doesn't report usage
- `cargo check --workspace` — clean
- `cargo clippy` — no new warnings


alecuba16:
Closing in favor of consolidated group branches. The changes from this PR have been merged into a functional group branch.

PR #375, opened 2026-08-02
fix(tui): hide tool content shows command but hides results

alecuba16:
## Summary

Fixes the `hide_tool_content` setting so that tool commands (`$ command`, `<> file paths`) are always shown, but tool result content is hidden with a "Result hidden" placeholder.

## Changes

- `render_tool_use_inner()` in `messages/mod.rs`: Removed `!hide_content` guard from bash `$ command` block so the command always renders
- Added `render_tool_result_hidden()` function returning a dim "Result hidden" placeholder
- Applied `hide_tool_content` guard to all four `ContentBlock::ToolResult` rendering paths
- Updated test `test_render_tool_use_hide_content` and added test `test_render_tool_result_hidden`

## Testing

- `cargo check --workspace` clean
- `cargo fmt --all` clean
- `cargo test --package claurst-tui` — all 5 tool rendering tests pass

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1255,3 +1255,44 @@ Some commands are available only under certain account or platform conditions:
 ### Feature-Flagged Commands
 
 Some commands check `isEnabled()` at runtime. For example, voice-related commands check for audio device availability; the desktop command checks for a display server.
+
+## Bang Commands (`!`)
+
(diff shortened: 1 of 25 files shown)


alecuba16:
Closing in favor of consolidated group branches. The changes from this PR have been merged into a functional group branch.

PR #370, opened 2026-08-01
feat: Cursor CLI ACP provider via Agent Client Protocol

alecuba16:
## Summary

Add Cursor CLI ACP (Agent Client Protocol) provider support, allowing Claurst to connect to Cursor as an LLM backend via the ACP subprocess protocol.

## Changes

- **New provider** (`crates/api/src/providers/cursor_acp.rs`): `CursorAcpProvider` implementing `LlmProvider` — request shaping, response parsing, streaming SSE handling, and tool-call conversion via the ACP wire format.
- **Provider registration** (`registry.rs`): `CursorAcpProvider` registered under the `"cursor-acp"` key in `provider_from_key`, `provider_from_config`, and `runtime_provider_for`.
- **Module exports** (`providers/mod.rs`): `pub mod cursor_acp` + re-exports of `CursorAcpCommand`, `CursorAcpProvider`, `ModelCatalog`.
- **TUI integration** (`app.rs`): provider picker entry, known providers list, and `activate_provider` match arm for cursor-acp subprocess management.
- **Query integration** (`query/src/lib.rs`): `"cursor-acp"` added to the native providers list.
- **Documentation** (`docs/providers.md`): setup instructions, environment variables, and `settings.json` configuration example.
- **Design doc** (`FEATURE_DESIGN_cursor_cli_acp_support.md`): architecture overview and protocol details.

## Setup

1. Ensure Cursor CLI is installed and available on `$PATH`.
2. Add to `settings.json`:
   ```json
   {
     "providers": {
       "cursor-acp": {}
     }
   }
   ```

## Test plan

- [ ] `cargo check --workspace` passes (verified locally)
- [ ] Provider appears in the model picker
- [ ] Basic prompt round-trip via Cursor ACP subprocess
- [ ] Streaming responses render correctly
- [ ] Tool-call round-trip works end-to-end

PR #366, opened 2026-08-01
feat: model picker overhaul — grouped providers, connected filter, Tab nav, favorites, default option, +6 fixes

alecuba16:
## Summary

Complete overhaul of the model picker: grouped by provider, connected-only filter, Tab navigation between groups, favorites, default option, cursor-acp model refresh, plus 6 bug fixes.

### Features (7)
- **Grouped by provider** — models grouped under provider name headers
- **Connected-only filter** — `a` key toggles showing only connected providers
- **Tab navigation** — Tab/BackTab jump between provider groups
- **Favorites** — `f` key star/unstar models, favorites sorted to top
- **Default option** — `Default` entry clears model override, persists across reboots
- **Cursor-acp model refresh** — triggers model discovery when picker opens in all-providers mode
- **Custom provider names** — group headers show registry provider names instead of "Other"

### Bug Fixes (6)
- Model picker all-providers mode picks correct provider (`filtered_models_grouped()` + navigation bounds)
- Capture `all_providers_mode` before `confirm()` closes picker
- Preselect correct row + match provider-prefixed model name
- Capture provider before close, skip `Other` header for default entry
- Picker group headers use registry provider names (`provider_names` HashMap)

### Files Changed (23 files)
- `src-rust/crates/tui/src/model_picker.rs` — `ModelPickerState` with `all_providers_mode`, `connected_provider_ids`, `show_all_providers`, `provider_names`, `filtered_models_grouped()`, Tab/BackTab, `a` toggle, favorites
- `src-rust/crates/tui/src/app.rs` — `is_provider_connected()`, `open_model_picker_all_providers()`, `set_model()` with `infer_provider_from_model()`
- `src-rust/crates/cli/src/main.rs` — cursor-acp model refresh, `is_current` skip in all-providers mode
- `src-rust/crates/tui/src/render.rs` — picker rendering, footer hints
- `docs/configuration.md` — default model docs
- `FEATURE_DESIGN_model_favorite_selector.md` — design doc
- 5 tests for preselection logic

### Note
Part of 5-PR set. Shares `app.rs`, `model_picker.rs`, `render.rs`, `main.rs` with other PRs. Consolidated from 13 branches (7 feat + 6 fix).

PR #365, opened 2026-08-01
feat: custom providers — multiple OpenAI-compatible, streaming settings, Cursor CLI ACP support

alecuba16:
## Summary

Adds support for multiple OpenAI-compatible custom providers and Cursor CLI ACP (Agent Client Protocol) provider.

### Features
- **Multiple custom providers** via `customProviders` array in `settings.json` — each with custom `apiBase`, `apiKey`, and model lists
- **Streaming settings** per provider: `streaming`, `includeUsageInStream`, `reasoningField`
- **Cursor CLI ACP provider** — connects to local Cursor subprocess via Agent Client Protocol
  - Session/prompt format with content blocks
  - ACP notification parsing
  - Model catalog discovery from session/new response
- Backwards compatible with single `customProvider` config

### Files Changed (21 files)
- `src-rust/crates/api/src/registry.rs` — `from_environment_with_auth_store()` reads `customProviders`
- `src-rust/crates/api/src/providers/cursor_acp.rs` — full ACP provider implementation
- `src-rust/crates/api/src/model_registry.rs` — `ProviderEntry` with custom provider support
- `src-rust/crates/core/src/lib.rs` — custom provider settings structs
- `src-rust/crates/tui/src/app.rs` — cursor-acp registration, `is_provider_connected()`, `open_model_picker_all_providers()`
- `src-rust/crates/tui/src/model_picker.rs` — picker support for custom providers
- `docs/providers.md`, `docs/configuration.md` — documentation
- `src-rust/crates/api/tests/test_custom_providers.rs` — tests

### Note
This PR is part of a 5-PR set grouped by area. PRs share some files (`app.rs`, `model_picker.rs`) and are designed to be reviewed together. Merge order: any order works, but shared files may require conflict resolution.

Consolidated from 3 branches (feat/multiple-openai-compatible-providers, feat/custom-provider-streaming-settings, feat/cursor-cli-acp-support).

PR #384, opened 2026-08-10
feat(tui): goal badge in transcript header + TODOs/branch in info panel

alecuba16:
Consolidates two feature branches:
- Goal badge in transcript header, simplified 'goal set' command
- TODOs label and git branch shown in info panel, removed duplicate from bottom bar

Supersedes #376 and #373.

PR #386, opened 2026-08-10
feat: custom providers and model picker overhaul

alecuba16:
Consolidates provider and model selection features:
- Custom providers: multiple OpenAI-compatible providers, streaming settings
- Model picker overhaul: grouped providers, connected filter, Tab nav, favorites, default option
- Cursor CLI ACP provider via Agent Client Protocol with reasoning effort support

**Cleaned up:** Removed shared payload from other group branches. No CRLF/rustfmt churn on untouched files (registry.rs went from 1258 to 6 lines of actual changes, providers/mod.rs from 87 to 3 lines).

- Picker keys (`f`, `*`, `a`) need to go through `crates/core/src/keybindings.rs`
- Tests need tempdir fixtures instead of real `~/.claurst/settings.json`
- Remove Revolut gateway URLs/model ids from docs and tests
- `Settings::load_sync()` should be resolved once into Config/registry
- Favourites matching needs provider-prefixed model id
- Cursor ACP should be a separate PR

Supersedes #365, #366, #370.


Kuberwastaken (changes requested):
Thanks for consolidating. Custom providers is a real gap on main and I want it, but this can't go in as one PR:

1. It still carries the todo-confidence, auto-poke, `!` bang-command, tok/s and keyboard-selection work that also lives in #384/#385/#387/#388 (shared commits 8674546/2d5d25d plus re-landed code). Reduce this PR to custom providers (+ the picker changes if you want them reviewed together) so the PRs are disjoint.
2. Picker: `f`, `*` and `a` are matched before `KeyCode::Char(c) => push_filter_char`, so you can no longer type "claude"/"haiku"/"flash" in the filter. New chords must go through `crates/core/src/keybindings.rs`. Also `/model` now always opens all-providers mode — behaviour change for everyone.
3. Tests must not read the developer's real `~/.claurst/settings.json` (`tests/test_custom_providers.rs`, the `provider_for_custom_id` tests, the query-crate env-var test). Use a tempdir fixture.
4. Remove the "from OpenCode config" section in docs/providers.md — it documents your internal Revolut gateway URLs and model ids, not the project. Same for the `revolut-ca/glm-5-2` ids in tests.
5. `Settings::load_sync()` is now on per-request and per-frame paths (`provider_for_id`, `is_openaiish_provider`, `build_provider_options`, `render_footer`'s `load_todos`). Resolve custom providers once into Config / the registry.
6. Favourites match on bare model id ignoring provider prefix, and the `f` handler keys off `config.provider` rather than the selected row's provider — wrong provider gets tagged in all-providers mode. A `customProviders.openai` entry silently shadows the real OpenAI provider; `/add` needs an id-collision check.
7. Cursor ACP: separate PR, and describe the model first. As written it runs `agent --force --trust`, drops the system prompt and all claurst tools/permissions, and buffers the whole turn before "streaming". I need to decide whether that belongs in claurst at all; if so, reuse `agent-client-protocol-schema` rather than raw JSON and don't register it unconditionally.

Will review the custom-providers piece on its own once the above is addressed.

alecuba16:
Thanks for the review. Splitting now.

1. **Scope:** Stripping the shared payload. This branch will contain only custom providers + model picker changes. Cursor ACP will go in a separate PR.
2. **Picker keys:** Will route `f`, `*`, `a` through `crates/core/src/keybindings.rs` so they don't shadow the filter char input. Will make `/model` keep its current behavior unless explicitly opened in all-providers mode.
3. **Tests:** Will switch all tests to tempdir fixtures. No reads of real `~/.claurst/settings.json`.
4. **Docs:** Will remove the "from OpenCode config" section and the `revolut-ca/glm-5-2` ids from docs and tests.
5. **Settings::load_sync():** Will resolve custom providers once into Config/registry instead of per-request/per-frame calls.
6. **Favourites:** Will fix matching to use provider-prefixed model id. Will fix the `f` handler to key off the selected row's provider. Will add an id-collision check for `customProviders.openai`.
7. **Cursor ACP:** Will be a separate PR with a proper description. Will look into `agent-client-protocol-schema` and won't register it unconditionally.

Will push the cleaned-up branches once ready.

Kuberwastaken (changes requested):
Thanks for sticking with this — and for the honest "known remaining issues" list in the description. Reading the current head (`3b37358e`), though, none of the seven items from my last review have landed yet, so I think this is still waiting on the push you mentioned rather than on new feedback from me. Flagging that first so we don't talk past each other.

A few things I found on this pass that weren't in my earlier list:

**Scope, measured.** I diffed this against #387 and #388: 14 of the 17 files here are *byte-identical* to their counterparts in both, and #386's added lines are 100% contained in #388 and 98.7% in #387 (`todo_write.rs`'s added block hashes identically across all three). The "Cleaned up: Removed shared payload" note in the description doesn't match what's on the branch — this PR still defines `BangCommandsConfig`, the `DisplayMessage::Bang*` variants and the `TodoItem` confidence fields, which are #388's feature. Whichever of the three lands first introduces those symbols and the other two become conflicting redefinitions, so I can't take any of them until they're genuinely disjoint.

**`execute_bang_command` needs to come out of this PR specifically.** It runs `bash -c <input>` gated only on `bangCommands.enabled` and plan mode — it doesn't go through the Bash tool's permission rules or the approval dialog, it's a blocking `.output()` on the render thread with no timeout, and it hardcodes `bash` so it breaks on Windows. That's a change I want to review on its own terms with that framing, not inside a providers PR.

**Three of the new `customProviders` knobs don't do anything yet.** `streaming` sets `capabilities().streaming`, but nothing in the workspace reads `caps.streaming` — `query/src/lib.rs:931` only consumes `image_input`/`tool_calling`/`thinking`, so the "falls back to non-streaming requests" comment describes behaviour that doesn't exist. `requestTimeoutSecs` is parsed and tested but never applied (`OpenAiCompatProvider` still uses the global `crate::request_timeout()`). `variants`/`ModelVariant` is parsed and tested but never consumed. Either wire them up or drop them from the schema for now — I'd rather ship a smaller surface that works.

**Two API-key paths that disagree.** `provider_for_custom_id` uses the new `resolve_api_key()` (returns `None` when the env var is unset), while `effective_config()` copies the raw value into `provider_configs` where `Config::resolve_provider_api_key` (`core/src/lib.rs:1492-1519`) runs the existing `substitute_env_vars` and yields `Some("")`. With `{env:FOO}` unset, the picker shows the provider as connected and the request 401s. Simplest fix is for `resolve_api_key` to just call `substitute_env_vars` rather than reimplement it — that also gets you embedded patterns like `Bearer {env:FOO}` for free.

**Smaller correctness bits:** `build_provider_options` lowercases `model_id` and then looks it up in `cp.models`, so any uppercase key in settings never matches its `reasoningEffort`. `apply_custom_providers` uses `context_window.unwrap_or(0)`, which `MIN_PLAUSIBLE_REGISTRY_WINDOW` then rejects, silently giving 100k — worth defaulting explicitly and warning. `valid_favorites` splits on the first `/`, so `free/auto` never matches; `ModelEntry` now carries `provider_id`, so use it for the match. And the `@@ -947,17 +947,6 @@` hunk in `core/src/lib.rs` accidentally eats the `ModelOverride` section header plus the first eight lines of its doc comment — it compiles, but the comment now starts mid-sentence. That one's identical in #387 and #388, so it looks like a de-dup pass artifact rather than intentional.

**On Cursor ACP** — I'd still like it as its own PR, and it does need to go through `crates/acp` / `agent-client-protocol-schema` rather than raw `serde_json::Value` (this PR adds no dependency at all). Three things I'd want fixed there regardless: the 8 KB stderr trim does `buf[start..]`, which panics on a UTF-8 boundary; `configured()` calls blocking `std::process::Command::status()` from `async fn health_check()` with no timeout on an executable named by an env var, and `registry.rs:366` reaches it from the status sweep; and `self.process` is never reset to `None`, so the "lazy reconnection" in the comment never happens.

Also still outstanding from last time: the tests reading the real `~/.claurst/settings.json` (`api/tests/test_custom_providers.rs` asserts `!models.is_empty()` against whatever the developer happens to have configured), the `revolut-ca/glm-5-2` and `llmg-coding` ids in tests, the `f`/`*`/`a` keys shadowing the `/model` filter, and `Settings::load_sync()` on per-request and per-frame paths — I counted about 24 new call sites, including `provider_for_custom_id`, `build_provider_options` and `provider_group_name`.

One genuinely nice thing to note: no hardcoded URLs, no key logging, no telemetry, no workflow or manifest changes — the security surface of the provider plumbing itself is clean.

Sequencing on my side: I'm landing #403 (the app.rs → `app/` module split) now, since all three of these need a restack anyway — so when you split, please branch from the new main. The decomposition I can review fastest: this PR reduced to custom providers + picker only, then bang/yolo/poke as separate small PRs (see my notes on #388), then the session browser work. Custom providers is the piece I actually want on main — happy to review it quickly once it's on its own.

alecuba16:
All seven review points are addressed in `dc26872`. Detail:

**1. Scope — custom providers + picker only.** Stripped everything else. No `BangCommandsConfig`, no `DisplayMessage::Bang*`, no `TodoItem` confidence fields, no auto-poke, no `!` bang-command, no keyboard-selection, no Cursor ACP. Diff is 15 files, all in the provider/picker path. `execute_bang_command` is gone. No `crates/acp` changes.

**2. Picker keys `f`/`a` routed through keybindings.** Added `KeyContext::ModelPicker` with 12 default bindings in `crates/core/src/keybindings.rs` (`prevModel`, `nextModel`, `pageUp/Down`, `firstModel`, `lastModel`, `selectModel`, `cancelModelPicker`, `effortDown/Up`, `nextProvider`, `prevProvider`, `toggleFavorite`, `toggleShowAllProviders`). `handle_model_picker_action` in `keys.rs` routes all picker keys through the configurable keybinding manager — the hardcoded `match key.code` block is gone. `f` and `a` are now proper keybinding actions, not inline `KeyCode::Char` matches, so typing "claude"/"haiku"/"flash" in the filter works. `/model` keeps current single-provider behavior (unchanged). `openModelPicker` (Ctrl+Shift+A) opens all-providers mode only when `customProviders` is non-empty.

**3. Tests use tempdir, not real `~/.claurst`.** Core tests (`crates/core/src/lib.rs`) build `Settings` in memory and call `effective_config()` directly — no filesystem access, no `Settings::load_sync()`, no `~/.claurst` reads. API tests (`crates/api/src/model_registry.rs`) use `std::env::temp_dir()` for the catalog cache file. No `revolut-ca/glm-5-2` or `llmg-coding` ids — test fixtures use `acme-gateway` and `My-Model-X`.

**4. No internal gateway docs.** `docs/providers.md` was not modified in this PR. No Revolut URLs, no internal model ids anywhere in the diff.

**5. `Settings::load_sync()` removed from per-request and per-frame paths.** Custom providers are resolved once into `Config.custom_providers` at load time (`effective_config()`). The picker reads from `Config`, not `Settings`: `load_favorites` takes a `Vec<String>` from `config.favorite_models`; `provider_group_name` reads from the cached `provider_names` map populated when the picker opens. `is_openaiish_provider` and `build_provider_options` take `&HashMap<String, CustomProviderDef>` from `config.custom_providers` — an in-memory lookup, not a settings read. Zero new `Settings::load_sync()` call sites in the per-request or per-frame paths.

**6. Favorites and id-collision fixes.** `valid_favorites` uses `ModelEntry.provider_id` for canonical key matching — `free/auto` and `openai/gpt-4o` are now distinguishable (no splitting on first `/`). `filtered_models_grouped` uses the same `is_fav` closure. The `f` handler in `handle_model_picker_action` reads the selected row's `provider_id` (from `filtered_models_grouped().get(selected_idx)`), not `config.provider` — correct provider gets tagged in all-providers mode. Id-collision check: `ProviderId::builtin_provider_ids()` returns all 47 builtin ids; `effective_config()` skips custom providers whose id collides with a builtin, logs a `tracing::warn!`, and excludes them from `Config.custom_providers`.

**Additional items from second review:**

- `streaming` and `variants`/`ModelVariant` fields dropped from the schema. Doc comment on `CustomProviderDef` explicitly states: "Streaming is always on — the workspace has no non-streaming fallback today, so a `streaming: false` knob would be dead configuration."
- `requestTimeoutSecs` is now applied: `OpenAiCompatProvider::with_request_timeout()` rebuilds the reqwest client with the per-provider timeout. Wired in `registry.rs` `provider_from_config`.
- `resolve_api_key()` calls `substitute_env_vars` (not a reimplementation), so `Bearer {env:FOO}` patterns work. Returns `None` when the key is absent or resolves to empty.
- `resolve_provider_api_key` filters empty AFTER `{env:VAR}` substitution — an unset env var yields `None`, not `Some("")`, so the picker no longer marks the provider connected.
- `build_provider_options` keeps exact-case `model_id_exact` for custom-provider `reasoningEffort` lookup (lowercased `model_id` is still used for the substring heuristics).
- `contextWindow` defaults to 128k with a `tracing::warn!` when unset (not `unwrap_or(0)` which hit `MIN_PLAUSIBLE_REGISTRY_WINDOW` rejection).
- `ModelOverride` doc comment is intact (the `@@ -947` hunk artifact is gone — verified section header and full doc comment at `core/src/lib.rs:1108`).
- Stale custom-provider cleanup drops only previously-custom providers (tracked via `old_custom` vs `new_custom` in `register_custom_providers`), never catalog-loaded providers.

**Tests:** `cargo check --workspace` clean. `cargo clippy --workspace --all-targets` clean. `cargo test -p claurst-core -p claurst-api -p claurst-tui -p claurst-query --lib` — 1447 passed, 0 failed (1 pre-existing failure in `settings_screen::tests::all_entries_returns_expected_settings` at HEAD, unrelated to this PR). 4 core tests, 3 api tests, 55 picker tests all pass.

alecuba16:
This PR had merge conflicts with main due to the `app.rs` → `app/` module split (#403). I've rebased the custom provider and model picker changes onto current main and opened #410 as a replacement.

The rebase maps `app.rs` changes to the new `app/` module files, excludes Cursor ACP (separate PR), bang command payload (#404), and other unrelated features. CRLF line endings preserved.

Closing in favor of #410.

Issue #382, opened 2026-08-05
Sessions are not working


mchaliadzinau:
Issue: `/session` always opens empty sessions window regardless of what I'm doing  `/fork`, `/new`.

`~/.claurst/sessions` contains session files

MacOS 26

ccfontes:
Same issue here, Claurst 0.1.7 on macOS 26. `/session` (and `/session list`, `/resume`) all return empty.

Diagnosed against the source at HEAD:

- `claurst_core::history::list_sessions()` in `src-rust/crates/commands/src/session.rs` (line 67) reads from `~/.claurst/projects/{base64url(project_root)}/{session_id}.jsonl` (per the comment in `src-rust/crates/core/src/session_storage.rs`).
- In my `~/.claurst/`, there is **no `projects/` directory at all** — only `sessions/`, `sessions.db`, `goals.sqlite`, `settings.json`.
- `~/.claurst/sessions/` contains 5 JSON files from sessions I ran today (timestamps match my use), so writes *are* happening, just to a path that `list_sessions` doesn't walk.
- The expected `projects/{base64url(cwd)}/` subdir does not exist on disk for my project root.

So: write path is `sessions/`, read path is `projects/{base64url(cwd)}/`. Either the read path constant moved between versions and the write path didn't follow, or the project-keyed subdirectory is never being created on first write.

ccfontes:
sorry guys, my AI did a "same here" post, only for me to find out later..

PR #377, opened 2026-08-02
feat(tui): add show-all paths and preview toggles to session browser

alecuba16:
## Changes

- Added `working_dir: Option<String>` to `SessionEntry`, populated from `ConversationSession.working_dir`
- Added `show_all` and `show_preview` toggles to `SessionBrowserState`
- **`a` key** — toggles showing each session's working directory path (extra indented line per row)
- **`p` key** — toggles preview panel showing selected session's ID, Title, Path, Updated, Msgs, Cost
- Hint bar shows both shortcuts with green/cyan on/off indicators
- Modal grows from 20 to 28 rows when preview is on

## Testing
- 4 new tests added (toggle_show_all_flips, toggle_show_preview_flips, close_resets_toggles, render_with_toggles_does_not_panic)
- `cargo check --workspace` clean
- `cargo clippy` clean
- All tests pass

PR #369, opened 2026-08-01
feat: keyboard text selection + copy in transcript

alecuba16:
## Summary

Keyboard text selection and copy in the transcript view.

### Features
- Shift+arrows for selection in transcript
- Cmd+C to copy selected text to clipboard
- Mouse select also works in error modals
- Selection state in `App` with start/end positions

### Files Changed (9 files)
- `src-rust/crates/tui/src/app.rs` — selection state + copy handler
- `src-rust/crates/tui/src/render.rs` — selection rendering (highlight selected text)

### Note
Part of 5-PR set. Shares `app.rs`, `render.rs` with other PRs. Consolidated from 1 branch.

PR #387, opened 2026-08-10
feat(tui): session browser improvements + keyboard text selection and copy

alecuba16:
Consolidates TUI interaction features:
- Session browser: show-all paths and preview toggles
- Keyboard text selection + copy in transcript (click-drag and keyboard)

**Cleaned up:** Stacked on top of #386 (providers-and-models). Incremental diff is only session_browser.rs + app.rs + input.rs + docs.

- `v` key needs dedicated entry binding in keybindings.rs (not bare `v` on empty prompt)
- Keyboard selection needs content-anchored instead of screen-row-anchored
- Docs need to match actual behavior
- `a`/`p` keys in session browser need to go through `crates/core/src/keybindings.rs`

Supersedes #377, #369.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1255,3 +1255,28 @@ Some commands are available only under certain account or platform conditions:
 ### Feature-Flagged Commands
 
 Some commands check `isEnabled()` at runtime. For example, voice-related commands check for audio device availability; the desktop command checks for a display server.
+
+## Keyboard Text Selection and Copy
+
+Use keyboard shortcuts to select text in the transcript and copy it to the
+clipboard. This works alongside the existing mouse-based click-drag selection.
+
+- **Arrow keys / hjkl**: Move the selection cursor in the transcript
+- **Shift+Arrow / Shift+hjkl**: Extend the selection
+- **Enter** or **`y`**: Yank (copy) the selected text to the system clipboard
+- **Esc**: Cancel the selection
+
+When mouse capture is enabled (default), click-drag selection also works.
+Set `"mouseCapture": false` in settings to release the mouse to the terminal
+for native selection instead.
+
+## Session Browser Improvements
+
+The session browser (`/session`) has been enhanced with:
+
+- **Show-all paths**: Toggle to reveal sessions from all projects, not just
+  the current one
+- **Preview toggles**: Quick toggle to show/hide session previews inline
+
+Use `/session` to open the browser, then follow the on-screen key hints to
+toggle these options.
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -767,3 +767,29 @@ When the keyed model exists in the catalog, the override patches it in place.
 When it does not (a self-hosted alias), Claurst materialises a synthetic entry
 so the corrected values flow everywhere the metadata is read: the `/model`
 picker, the token-usage warnings, and the auto-compact thresholds.
+
+## Cursor ACP (`cursor-acp`)
+
+Connect to Cursor's CLI agent via the Agent Client Protocol (ACP).
+
+### Setup
+
+1. Install Cursor CLI (`agent` must be on PATH)
+2. Set `cursor-acp` as your provider:
+```json
+{
+  "provider": "cursor-acp"
+}
+```
+
+### Environment Variables
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `CLAURST_CURSOR_ACP_PATH` | `agent` | Path to the Cursor CLI executable |
+| `CLAURST_CURSOR_ACP_ARGS` | `acp` | ACP subcommand argument |
+| `CLAURST_CURSOR_ACP_EXTRA_ARGS` | `--force --trust` | Permission arguments before `acp` |
+| `CLAURST_CURSOR_ACP_MODEL` | (empty) | Default model ID |
+
+The provider spawns `agent --force --trust acp` as a subprocess and
+communicates via newline-delimited JSON-RPC 2.0 over stdin/stdout.
--- a/src-rust/crates/api/src/model_registry.rs
+++ b/src-rust/crates/api/src/model_registry.rs
@@ -1243,6 +1243,66 @@ impl ModelRegistry {
         }
         apply_overrides_to_entries(&mut self.entries, &self.overrides);
     }
+
+    /// Register user-defined custom provider models into the registry.
+    /// Each model is keyed as `<provider_id>/<model_id>` and a synthetic
+    /// `ModelEntry` is created from the `CustomModelDef` metadata.
+    /// A `ProviderEntry` is also created for each custom provider so it
+    /// appears in `/providers` listing.
+    pub fn apply_custom_providers(
+        &mut self,
+        custom_providers: &HashMap<String, claurst_core::config::CustomProviderDef>,
+    ) {
+        for (provider_id, def) in custom_providers {
+            // Register the provider entry so it shows up in /providers.
+            self.providers.entry(provider_id.clone()).or_insert(ProviderEntry {
+                id: ProviderId::new(provider_id),
+                name: def.name.clone(),
+                env: Vec::new(),
+                api: Some(def.api_base.clone()),
+                npm: None,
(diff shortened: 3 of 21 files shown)


Kuberwastaken (changes requested):
Thanks, but this branch still carries the whole mega-branch payload: custom providers, model picker overhaul, bang commands, todo confidence/auto-poke and TPS are all in here and are the same changes as #384/#385/#386/#388. The session-browser toggles and keyboard selection are ~300 real lines; the rest of the +5384 is that payload plus CRLF->LF and rustfmt churn on app.rs / model_picker.rs / session_browser.rs / registry.rs.

Please split and rebase on main without the shared payload:
1. feat(tui): session browser path/preview toggles — fine as-is, but fix docs/commands.md ("show-all" does not list other projects; it shows each session's path) and route `a`/`p` through `crates/core/src/keybindings.rs`.
2. feat(tui): keyboard selection — blocking issues: bare `v` on an empty prompt eats the first character of any message starting with v (and shadows vim's `v`); all keys are inline KeyCode matches instead of keybindings.rs actions in `KeyContext::Transcript`; docs promise hjkl / Shift-extend / Enter-yank which don't exist; the Up/Down scroll branches are unreachable so selection can't leave the visible area. Needs a dedicated entry binding in keybindings.rs, content-anchored rather than screen-row-anchored selection, and docs that match.

Drop the rustfmt/CRLF rewrites from the diff — fmt is advisory here and main is intentionally not fmt-clean.

alecuba16:
Thanks for the review. Splitting and rebasing now.

1. **Session browser path/preview toggles:** Will fix docs/commands.md ("show-all" shows each session's path, not other projects) and route `a`/`p` through `crates/core/src/keybindings.rs`.
2. **Keyboard selection:** Will add a dedicated entry binding in keybindings.rs (no bare `v` on empty prompt), switch to content-anchored selection instead of screen-row-anchored, fix docs to match actual behavior (drop hjkl/Shift-extend/Enter-yank that don't exist), and fix the Up/Down scroll branches so selection can leave the visible area.

Dropping the rustfmt/CRLF rewrites from the diff.

Kuberwastaken (changes requested):
Thanks for rebasing onto main. Reading the current head (`1229edca`), though, the branch is still carrying the whole #386 payload — I diffed it file-by-file against #386 and **15 of the 19 files are byte-identical**: `cursor_acp.rs` (+1412), `model_picker.rs` (+964), `core/src/lib.rs` (+374), `todo_write.rs`, `render.rs`, all of `api/` and `query/`. Commit `3b37358e` is literally #386's commit (it's in #388 too). The 08-24 push rebased but didn't split, and none of the six items from the last review landed. Re-flagging them so nothing gets lost:

**Still open from last time**
1. **Split the branch.** Only `1229edca` belongs here (+1983/-1129 across 4 files). Rebase onto main with just that.
2. **Drop the fmt/CRLF rewrites.** `session_browser.rs` is one `@@ -1,605 +1,838 @@` whole-file hunk — CRLF→LF plus rustfmt around ~200 real lines. Same story for the `-523` in `app.rs`. `ci.yml:68` explains why the tree is intentionally not fmt-clean.
3. **Bare `v` on an empty prompt** still eats the first character of "verify …" and shadows vim's Normal-mode `v` — the new arm sits right before the generic `KeyCode::Char(c)` arm that routes to `vim_command`. Needs a dedicated binding in `keybindings.rs`.
4. **`KeyContext::Transcript` still bypassed** — `crates/core/src/keybindings.rs:277-284` already has up/down/pageup/pagedown/home/end/enter/escape defined; the new keys are hardcoded `KeyCode` matches.
5. **The Up/Down scroll branches are still unreachable.** `Up` guards `kb_cursor_row > area.y` then tests `kb_cursor_row < area.y` after decrementing; `Down` guards `< max_row` (`= area.y+height-1`) then tests `>= area.y+height`. Neither can fire, so selection can't leave the visible area.
6. **The docs still describe a different feature** — `hjkl`, `Shift+Arrow` extend and `Enter`-to-yank don't exist (`Enter` is intercepted by `run_interactive`'s `plain_enter` branch before `handle_key_event` sees it), `v` is never mentioned even though it's the only way in, and "show-all reveals sessions from all projects" isn't what the code does — it prints each already-listed session's `working_dir`.

**New, from this pass**
7. **The session-browser half never runs.** `working_dir: s.working_dir` goes into `App::run` (`app.rs:6484`), which has no callers — the real loop is `run_interactive` in `crates/cli/src/main.rs`. `session_list_pending` is set by `/session` at `app.rs:2097` and consumed only inside `App::run`; `grep` finds no reference to it anywhere in `crates/cli`. At runtime the browser opens empty (that's open issue #382), so `a` toggles paths on nothing and `p` shows "(no session selected)". Populating the list from the real loop is the actual prerequisite here — I'm fixing #382 on main, so rebase and build on that.
8. **`Home` scrolls to the bottom.** `scroll_offset = self.total_message_lines.get().saturating_sub(area.height)` — but `total_message_lines` is initialized to `0` and written nowhere in the workspace. Your line is its only reader, so it's always `0.saturating_sub(h) == 0`.
9. **`last_selectable_area` is the whole terminal**, not the transcript — `render.rs:537` sets it to `frame.area()`. The cursor roams over the prompt box and status bar, and copy picks up UI chrome. This is the concrete reason to go content-anchored.
10. **No visible cursor on entry.** `update_kb_selection` pins both anchor and focus to column `area.x`, and `apply_selection_highlight` (`render.rs:928`) early-returns when `anchor == focus` — press `v` and nothing highlights.
11. **`_ => return false`** swallows Ctrl+D, so you can't exit while in selection mode.
12. **The preview panel gets clipped.** `render_session_browser` builds one flat `Vec<Line>` and renders it through `Paragraph` with no `.scroll()` into a fixed 20/28-row modal, with the preview and hint bar appended *after* the list. Past ~14 sessions neither is on screen — and `show_all` halves that. Worth windowing the list around `selected_idx` while you're in there.
13. **Test coverage.** `kb_select_mode_starts_and_cancels` sets the fields by hand and asserts `clear_selection()` clears them — it never goes through `handle_key_event`. A test that drives real key events would have caught #5 and #8.

**Sequencing:** I'm landing #403 (the app.rs → `app/` module split) now, since all of the group PRs need a restack anyway — so when you split this out, branch from the new main and the selection block will target `app/keys.rs`. Also note the `render.rs`/`todo_write.rs` hunks currently here duplicate #397 (now closed — see my reasoning there), and `render_bang_display_lines` + `DisplayMessage::Bang*` duplicate #388.

Good news: CI ubuntu is green including `clippy -D warnings`, and reusing `image_paste::write_clipboard_text` for the copy path was the right call — no new dependencies anywhere in the PR. The selection feature itself is only ~130 lines; as a standalone PR against `keybindings.rs` with content-anchored coordinates it'd be a quick review.

alecuba16:
Splitting #387 into two PRs to keep scope tight:

- **#407 (this split A)**: keyboard selection, path/preview toggles for the session browser, and fix #382 (browser opened empty). All keys route through `keybindings.rs`. Tests: 5 unit + 2 TUI.
- **Split B (coming next)**: keyboard text selection + copy within the session browser.

No shared payload between the two.

alecuba16:
Split B of #387 is now up as #408: keyboard text selection + copy in transcript.

Addresses every item from the review:
- `v` on empty prompt enters selection mode (gated on no voice recorder, not vim Normal/Visual)
- All selection keys route through `KeyContext::Transcript` in keybindings.rs
- Content-anchored: `last_selectable_area` restricted to `last_msg_area`, `total_message_lines` written at render time
- Scroll branches reachable, cursor clamps to bounds
- Visible cursor on entry (anchor == focus shows cursor bar)
- Ctrl+D not swallowed (allowed to exit)
- 8 new tests driving real key events
- Docs match actual behavior

No shared payload with #407 (split A). Both are standalone against main.

alecuba16:
All review points addressed in the splits: #407 (session browser) and #408 (keyboard selection). Both mergeable on current main. Closing this in favor of the splits.

PR #378, opened 2026-08-02
feat: smart auto-poke with guardrail, error detection, /poke status

alecuba16:
### Independent from input flow
- Auto-poke submits directly to model via `Message::auto_poke()` + `run_query_loop`
- Does NOT touch `prompt_input.text` — user can keep typing while poke runs

### Smart features
- **Guardrail circuit breaker** — stops after 3 consecutive provider refusals
- **Non-retryable error detection** — stops on billing/auth/payload-size errors
- **`/poke on|off|status`** — subcommand variants with detailed status display (pokes used, budget left, incomplete todos, guardrail stops, blocked status)
- **User-visible messages** — "X incomplete todos. Poking. /poke off to stop." / "All todos done." / "Provider refused N turns — stopping."
- **Distinct dim yellow color** for auto-poke messages in transcript
- **Shorter prompt** — "X incomplete todos. Continue." instead of verbose original

## Files changed
- `crates/core/src/lib.rs` — `is_auto_poke` field on `Message`, `auto_poke()` constructor
- `crates/tui/src/app.rs` — guardrail/error fields, `check_auto_poke()` rewrite, `/poke` command
- `crates/cli/src/main.rs` — direct model submission path
- `crates/tui/src/messages/mod.rs` — auto-poke color constants, conditional rendering
- `crates/tools/src/todo_write.rs` — `is_non_retryable_error()`, `is_guardrail_error()`, shorter message

## Testing
- `cargo check --workspace` clean, `cargo fmt` clean
- 15 session browser tests, 19 todo_write tests pass

PR #374, opened 2026-08-02
feat: add persistent yolo mode setting

alecuba16:
## Summary

Adds persistent YOLO mode (auto-approve all tool calls without prompts) that survives restarts and is inherited by sub-agents.

## Changes

- **Settings**: `yoloMode: bool` in `settings.json`, serialized as `"yoloMode"`, merged via OR in `Settings::merge()`
- **effective_config()**: forces `BypassPermissions` when `yolo_mode` is true
- **CLI**: `--permission-mode` changed to `Option<CliPermissionMode>`, preserving settings value when unset; `--yolo`/`--dangerously-skip-permissions` overrides settings
- **TUI**: `/yolo` slash command toggles at runtime, persists to `settings.json`, updates `config.permission_mode` immediately
- **Settings screen**: YOLO mode toggle entry
- **Config tool**: `get`/`set` `yoloMode` via `/config`
- **Sub-agents**: inherit parent's `permission_mode` via cloned `ToolContext` — no extra wiring needed
- **Docs**: `configuration.md` documents `yoloMode` key

## Test plan

- `cargo test --package claurst-core -- yolo_mode` — 3 tests pass
- `cargo test --package claurst-tui -- test_yolo_slash_command` — passes
- `cargo check --workspace` — clean

PR #368, opened 2026-08-01
feat: bang (!) batch commands — shell exec bypassing model, zero tokens

alecuba16:
## Summary

Adds bang (`!`) batch command support — commands prefixed with `!` execute directly in the shell, bypassing the model entirely (zero tokens consumed).

### Features
- `! <command>` → runs shell command, shows output in transcript
- `!! <command>` → multi-line mode (pastes output into input buffer)
- Works in plan mode (blocked with warning)
- Output rendered via `render_bang_display_lines()`
- Intercept in `main.rs` main loop (was dead code in `App::run`)
- Render output when transcript empty (`transcript_empty` includes `display_messages.is_empty()`)

### Files Changed (18 files)
- `src-rust/crates/cli/src/main.rs` — bang intercept in main loop
- `src-rust/crates/tui/src/render.rs` — `render_bang_display_lines()` + bang output rendering
- `src-rust/crates/tui/src/app.rs` — bang command state handling
- `src-rust/crates/tui/src/input.rs` — bang command input parsing
- `src-rust/crates/core/src/lib.rs` — `bangCommands.enabled` setting
- `docs/commands.md` — documentation
- `FEATURE_DESIGN_batch_commands_no_model.md` — design doc

### Note
Part of 5-PR set. Shares `app.rs`, `render.rs`, `main.rs` with other PRs. Consolidated from 2 branches (1 feat + 1 fix).

PR #388, opened 2026-08-10
feat: bang batch commands, yolo mode, /poke auto-poke command

alecuba16:
Consolidates command and settings features:
- Bang (!) batch commands: shell exec bypassing model, zero tokens
- Persistent yolo mode setting (yoloMode in settings.json, /yolo toggle)
- /poke slash command to toggle/configure auto-poke (on/off/status)

**Cleaned up:** Stacked on top of #386 (providers-and-models). Incremental diff is 501 lines across 8 files (bang + yolo + poke only). No shared payload.

- Security: `yoloMode` must be global-only with root/sudo guard
- Bang exec needs spawned task with streamed output instead of synchronous `bash -c`
- Auto-poke should be opt-in (default OFF)
- Keybindings need to go through `crates/core/src/keybindings.rs`

Supersedes #368, #374, #378.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1255,3 +1255,44 @@ Some commands are available only under certain account or platform conditions:
 ### Feature-Flagged Commands
 
 Some commands check `isEnabled()` at runtime. For example, voice-related commands check for audio device availability; the desktop command checks for a display server.
+
+## Bang Commands (`!`)
+
+Execute shell commands directly from the input prompt without going through
+the model. Zero token consumption. Output is display-only — the model never
+sees it.
+
+### Usage
+
+```
+! ls -la
+! git status
+! echo "hello"
+```
+
+Type `!` followed by a shell command. The command runs via `bash -c` in the
+project working directory and the result is shown inline in the transcript.
+
+### Configuration (`settings.json`)
+
+```json
+"bangCommands": {
+  "enabled": true,
+  "addToHistory": true,
+  "showInTranscript": true
+}
+```
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `enabled` | `false` (opt-in) | Toggle the feature |
+| `addToHistory` | `false` | Store `!` commands in a separate shell-command history |
+| `showInTranscript` | `true` | Display command + output in the chat transcript |
+
+### Notes
+
+- Output is display-only — the model never sees it (zero token consumption).
+- Blocked in plan mode (read-only restriction).
+- Uses `bash -c` for execution (not the PTY-based Bash tool).
+- `!!` (double bang) is NOT treated as a bang command.
+- A single `!` with no command is ignored.
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1255,3 +1255,75 @@ Some commands are available only under certain account or platform conditions:
 ### Feature-Flagged Commands
 
 Some commands check `isEnabled()` at runtime. For example, voice-related commands check for audio device availability; the desktop command checks for a display server.
+
+## Bang Commands (`!`)
+
+Execute shell commands directly from the input prompt without going through
+the model. Zero token consumption. Output is display-only — the model never
+sees it.
+
+### Usage
+
+```
+! ls -la
+! git status
+! echo "hello"
+```
+
+Type `!` followed by a shell command. The command runs via `bash -c` in the
+project working directory and the result is shown inline in the transcript.
+
+### Configuration (`settings.json`)
+
+```json
+"bangCommands": {
+  "enabled": true,
+  "addToHistory": true,
+  "showInTranscript": true
+}
+```
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `enabled` | `false` (opt-in) | Toggle the feature |
+| `addToHistory` | `false` | Store `!` commands in a separate shell-command history |
+| `showInTranscript` | `true` | Display command + output in the chat transcript |
+
+### Notes
+
+- Output is display-only — the model never sees it (zero token consumption).
+- Blocked in plan mode (read-only restriction).
+- Uses `bash -c` for execution (not the PTY-based Bash tool).
+- `!!` (double bang) is NOT treated as a bang command.
+- A single `!` with no command is ignored.
+
+## `/yolo` — Toggle YOLO Mode
+
+Toggle YOLO mode, which skips all permission prompts and auto-approves every
+tool call. When enabled, the effective permission mode is set to
+`bypassPermissions`. The setting persists in `settings.json` under
+`"yoloMode"` and is restored on the next session.
+
+```
+/yolo          — toggle YOLO mode on/off
+```
+
+YOLO mode propagates to sub-agents: they inherit the parent's permission mode
+via the cloned tool context. You can also enable YOLO for a single session via
+the CLI flag `--dangerously-skip-permissions` (alias `--yolo`).
+
+## `/poke` — Toggle Auto-Poke
+
+Toggle or configure the auto-poke feature, which automatically sends a
+continuation prompt when the model stops with incomplete todos.
+
+```
+/poke          — toggle auto-poke on/off
+/poke on       — enable auto-poke
+/poke off      — disable auto-poke
+/poke status   — show current status, budget, and incomplete todo count
+```
+
+Auto-poke has a safety budget of 48 pokes per session and stops after 3
+consecutive no-progress turns. The setting persists in `settings.json` under
+`"autoPokeEnabled"` (default: `true`).
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -42,6 +42,7 @@ values. Keys absent from the project file fall back to the global value.
   "config": { ... },
   "providers": { ... },
   "modelOverrides": { ... },
+  "favoriteModels": [ ... ],
   "projects": { ... },
   "commands": { ... },
   "formatter": { ... },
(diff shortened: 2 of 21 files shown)


Kuberwastaken (changes requested):
Thanks for consolidating, but this can't go in as is.

1. Security: `yoloMode` is read from the merged (global + project) settings at main.rs:550 and `||`-merged, so any repo's `.claurst/settings.json` can flip bypassPermissions on, and the root/sudo guard is skipped. Must be global-only and keep the guard. Also `yolo_command_toggles_permission_mode` writes `yoloMode: true` into the developer's real settings.json when running `cargo test`.
2. Scope: ~3000 of the 3666 lines are customProviders / model picker / favorites / confidence / tok-s / keyboard selection / CRLF rewrites that also live in #384–#387. Rebase onto main with only bang + yolo + poke, one commit each.
3. Bang exec: synchronous `bash -c … .output()` in the event loop blocks the TUI with no timeout/cancel and hardcodes bash (Windows is supported); output is appended at the transcript tail instead of in sequence; `!` is intercepted in both main.rs and app.rs; bypasses the permission system. Should go through the existing Bash/PTY tool plumbing or a spawned task with streamed output.
4. Auto-poke: default ON with a 48-poke budget and no stop-reason check will auto-continue through the model's clarifying questions and burn tokens; make it opt-in, drop the duplicate `/todos auto-poke` toggle, remove unused fields (`pending_auto_poke`, `todo_card_visible`, `TodoCard`, `bang_command_history`).
5. `v` selection-mode key and picker keys are hardcoded — route through `crates/core/src/keybindings.rs`. Drop per-frame `load_todos` in `render_footer` and the ad-hoc `Settings::load_sync()` calls (20 sites vs 8 on main).

Happy to review a trimmed PR.

alecuba16:
Thanks for the detailed review. Working through each point:

1. **Security (yoloMode):** Will make `yoloMode` global-only (read from `Settings::load_global()`, not merged config) and keep the root/sudo guard. Will also fix the test so it writes to a tempdir, not real `settings.json`.
2. **Scope:** Rebasing now to strip the shared payload. This branch will contain only bang + yolo + poke, one commit each.
3. **Bang exec:** Will move to spawned task with streamed output instead of synchronous `bash -c`. Will go through the existing tool plumbing. Will also remove the duplicate `!` interception in app.rs.
4. **Auto-poke:** Will make it opt-in (default OFF), drop the duplicate `/todos auto-poke` toggle, and remove unused fields (`pending_auto_poke`, `todo_card_visible`, `TodoCard`, `bang_command_history`).
5. **Keybindings:** Will route `v` and picker keys through `crates/core/src/keybindings.rs`. Will drop per-frame `load_todos` in `render_footer` and the ad-hoc `Settings::load_sync()` calls.

Will push the cleaned-up branch once ready.

Kuberwastaken (changes requested):
Thanks for the rebase — but I checked the current head against the five points from my last review and none of them landed. Walking through them:

**1. yoloMode security — still unfixed, and it's the blocker.** `main.rs:~550` still reads `settings.config.yolo_mode` from `Settings::load_hierarchical(&cwd)`, and `merge()` still does `over.config.yolo_mode || base.config.yolo_mode`. Since `merge(base, over)` is called as `merge(global, project)`, `over` is the repo's `.claurst/settings.json` — so any cloned repo can ship `{"config":{"yoloMode":true}}` and get bypassPermissions on first launch, with no trust prompt and no way for a global `false` to override it. The root/sudo guard is explicitly skipped too, so `sudo claurst` in such a repo is root-level auto-approved exec. This is a third sink on top of #389, and a worse one — no prompt submit required.

The pattern I want is already in the file: `skip_dangerous_mode_permission_prompt: base.skip_dangerous_mode_permission_prompt` in `merge()`. Please pin `yolo_mode` (and `bangCommands.enabled`) to `base` the same way, read yolo from global settings only, and keep the root/sudo bail. Also: runtime `/yolo` should go through the `BypassPermissionsModeDialog` gate — right now that check is startup-only, so the toggle grants bypass with zero confirmation. And `/yolo off` shouldn't hard-set `Default`; that silently kicks the user out of `plan`/`acceptEdits`.

`yolo_command_toggles_permission_mode` still writes to the real `~/.claurst/settings.json` via `save_sync()`, and the last state it sets is `on` — so `cargo test` leaves YOLO enabled globally for whoever ran it. Needs a tempdir.

**2. Scope — went the wrong way.** The diff against `main` is now 4543 additions, up from 3666. Stacking on #386 makes the branch *smaller relative to #386*, but this PR merges into `main`, so `gh pr diff 388` is what lands: all of cursor_acp.rs, model_picker.rs, customProviders, favorites. And even the incremental slice still has `/add` (+108 in `commands/src/providers.rs`), `TodosCommand` (+63), and the favoriteModels/customProviders docs — all provider/model scope. What I'm asking for is three PRs branched off `main`, one feature each: bang, yolo, poke. Not a stack.

**3. Bang exec — unchanged.** Still `std::process::Command::new("bash").arg("-c")…output()` synchronously on the event-loop thread: no timeout, no cancel, no output cap (`! npm install` or `! yes` hangs or OOMs the TUI), and hardcoded `bash` on a repo that supports Windows. Still intercepted in both `main.rs` and the dead `App::run` in `app.rs` — drop the app.rs copy.

**4. Auto-poke — unchanged.** Still `default = "default_true"`, `/todos auto-poke` is still a duplicate toggle for `/poke`, and `pending_auto_poke` / `todo_card_visible` / `bang_command_history` / `TodoCard` are still dead. The stop-reason check is still missing: `check_auto_poke()` is called right after `complete_current_turn_snapshot(stop_reason.contains("abort") || …)` without consulting it, so hitting Esc to stop the model re-pokes it immediately. With yolo on that's an uninterruptible 48-turn auto-approving loop.

**5. Perf/keybindings — unchanged.** `load_todos()` (disk read + JSON parse) is still called inside `render_footer`, i.e. every frame. `Settings::load_sync()` sites went from 8 to 35 in `app.rs`. The `v` selection key and picker keys still bypass `crates/core/src/keybindings.rs`.

**Separately, and more urgently:** #386 and #387 both contain the *entire* bang-command and auto-poke implementations — `execute_bang_command`, `Command::new("bash")`, `check_auto_poke`, `autoPokeEnabled` default-true, all verbatim. So "custom providers, model picker overhaul, Cursor ACP support" would silently land a synchronous shell-exec path and a default-ON self-continuation loop. Please strip those out of #386 and #387 as well; I can't merge a provider PR that ships shell exec.

Merge order: I'm landing #403 (the app.rs → `app/` split) now since all three group PRs need a restack anyway — branch the three small PRs off the new main and the command hunks will target `app/commands.rs`.

Happy to review these as three small PRs off `main`. Bang commands with a spawned task + timeout + output cap is the one I'd merge first.

alecuba16:
Splitting #388 into three separate PRs as requested. The first one is up:

- **#404 — Bang commands** (this PR): spawned task + timeout + output cap, cross-platform, no hardcoded bash. The synchronous `bash -c .output()` is gone.

Coming next:
- **Yolo mode**: global-only read, keep root guard, tempdir test, runtime `/yolo` through `BypassPermissionsModeDialog`.
- **Auto-poke**: opt-in (default OFF), stop-reason check, drop duplicate `/todos auto-poke`, remove dead fields.

All three branch off main, no shared payload.

alecuba16:
Second split from #388 is up: **#405 — YOLO mode**.

Review points from this thread, addressed:

1. **yoloMode global-only + root guard** — `Settings::merge` pins `yolo_mode: base.config.yolo_mode` (same pattern as `skip_dangerous_mode_permission_prompt`), and the startup path keeps the `Uid::effective().is_root()` bail. A repo's `.claurst/settings.json` cannot enable or disable it.
2. **Runtime `/yolo` through `BypassPermissionsModeDialog`** — `/yolo on` shows the gate first; accept activates, decline cancels without exiting.
3. **`/yolo off` restores the previous mode** — remembered before activation, so plan/acceptEdits sessions are not kicked to Default.
4. **No test writes real settings** — the project-cannot-enable test uses tempdirs and an isolated `CLAURST_HOME`.

Third split (auto-poke: default OFF, stop-reason check, drop `/todos auto-poke` duplicate) coming next. All three branch off main with no shared payload.

alecuba16:
Third and final split from #388 is up: **#406 — auto-poke**.

Review points from this thread, addressed:

1. **Opt-in, default OFF** — `autoPokeEnabled` uses `#[serde(default)]`, not `default_true`. Absent key means disabled. Covered by a default-off roundtrip test.
2. **Stop-reason check (the blocker)** — `check_auto_poke(stop_reason)` is called with the actual stop reason from `TurnComplete` and only fires on a clean `end_turn`. Aborts/cancels (Esc), `max_tokens` truncation, `content_filtered`, and unknown stop reasons never poke. Hitting Esc to stop the model no longer re-pokes it. Test asserts all four abnormal reasons queue nothing.
3. **Duplicate `/todos auto-poke` dropped** — `/poke` is the single entry point. No `TodosCommand` auto-poke path exists in this PR.
4. **Dead fields removed** — `pending_auto_poke`, `todo_card_visible`, `bang_command_history`, and `TodoCard` from the original branch are not carried over. Only the three fields auto-poke actually uses exist: `auto_poke_count`, `no_progress_turns`, `last_todo_hash`.
5. **No render-loop disk reads** — `load_todos` runs once per completed turn inside the `TurnComplete` handler, never in `render_footer`.
6. **Safety rails kept** — budget of 48 pokes per session, stops after 3 consecutive no-progress turns (unchanged todo-state hash). Both have tests.

Also: `App.session_id` is synced with `ToolContext::session_id` at startup, `/new`, and session resume so pokes read the right session's todos. TUI tests mount a tempdir `CLAURST_HOME` so nothing touches real user state.

`cargo check --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, 2 core tests, 7 TUI tests, and the full claurst-tools suite pass. CRLF line endings preserved on `todo_write.rs` and `docs/commands.md`.

All three splits are now up: #404 (bang), #405 (yolo), and this one. Each branches off main with a single-feature diff.

alecuba16:
All review points addressed in the three splits: #404 (bang commands), #405 (YOLO mode), #406 (auto-poke). All mergeable on current main. Closing this in favor of the splits.

PR #398, opened 2026-08-24
feat(tui): add /turns slash command to set or disable max turn limit

alecuba16:
## Summary
Adds `/turns` slash command to set or disable the max turn limit for the current session.

- `/turns` shows current turn limit
- `/turns <N>` sets max turns to N
- `/turns off` disables turn limit
- Turn limit enforced in the CLI main loop

Clean branch from current main, no shared payload. 36 lines, 2 files.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -7,7 +7,7 @@ This document is the complete reference for every slash command available in Cla
 ## Table of Contents
 
 1. [Command System Overview](#command-system-overview)
-2. [Session & Navigation](#session--navigation)
+2. [Session & Navigation](#session--navigation) — `/help`, `/clear`, `/new`, `/move`, `/exit`, `/resume`, `/session`, `/fork`, `/rename`, `/rewind`, `/compact`, `/turns`
 3. [Model & Provider](#model--provider) — `/model`, `/providers`, `/connect`, `/thinking`, `/effort`, `/advisor`, `/fast`
 4. [Configuration & Settings](#configuration--settings) — `/config`, `/keybindings`, `/permissions`, `/hooks`, `/privacy-settings`, `/mcp`, `/output-style`, `/theme`, `/statusline`, `/vim`, `/voice`, `/terminal-setup`
 5. [Code & Git](#code--git) — `/commit`, `/diff`, `/undo`, `/review`, `/security-review`, `/init`, `/search`
@@ -186,6 +186,24 @@ Summarize and compress the conversation history to reduce context window usage. 
 ```
 /compact
 ```
+
+### /turns
+
+Show, set, or disable the session max-turn limit.
+
+```
+/turns          # show the current limit
+/turns 25       # cap this session at 25 turns
+/turns off      # raise the limit to a large finite ceiling (1000)
+```
+
+The override is **session-local**: it is not persisted to `settings.json` and
+resets to the config/agent default on restart. Precedence is override ->
+agent (`max_turns` in an agent definition) -> config (`max_turns`).
+`/turns off` does not remove the limit entirely -- it raises it to a finite
+ceiling of 1000 turns so runaway loops still terminate. When the cap is
+exceeded, the usual graceful degradation applies (see
+`degradationSummaryEnabled`).
 
 ---
 
--- a/src-rust/crates/cli/src/main.rs
+++ b/src-rust/crates/cli/src/main.rs
@@ -2988,6 +2988,9 @@ async fn run_interactive(
                         let tools_arc_clone = tools_arc.clone();
                         let mut ctx_clone = tool_ctx.clone();
                         let mut qcfg = base_query_config.clone();
+                        // Session /turns override applies at query-build time
+                        // (override → agent → config precedence, set via /turns).
+                        qcfg.max_turns_override = app.max_turns_override;
                         qcfg.model = claurst_api::effective_model_for_config(&cmd_ctx.config, &model_registry);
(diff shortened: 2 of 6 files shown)


Kuberwastaken (changes requested):
Thanks — a session-local turn cap override is a reasonable thing to want, and the command dispatch does work: `turns` isn't in the `skip_tui_for_args` list in `main.rs`, `find_command("turns")` returns `None` so `execute_command` doesn't fire, and `handled_by_tui` suppresses the "Unknown command" fallback. So `/turns`, `/turns 25` and `/turns off` all reach your handler.

The problem is that the override doesn't reach the turn cap that actually applies. `run_query_loop` computes:

```rust
// crates/query/src/lib.rs:430
let effective_max_turns = config.agent_definition
    .as_ref()
    .and_then(|a| a.max_turns)
    .unwrap_or(config.max_turns);
```

The agent definition wins over `config.max_turns`. The built-in `plan` agent sets `max_turns: Some(20)` (`crates/core/src/lib.rs:1380`), so in plan mode `/turns 50` and `/turns off` are silently ignored and the loop still stops at 20 — the opposite of what the new comment ("takes precedence over agent defaults") promises. Setting `base_query_config.max_turns` only helps when no agent definition carries a `max_turns`.

The fix needs to be at the `effective_max_turns` site — e.g. an explicit `max_turns_override: Option<u32>` on `QueryConfig` consulted first, so the precedence reads override → agent → config.

A few smaller points:

- **`off` maps to `u32::MAX`,** which means a genuinely unbounded loop with no cap *and* no degradation summary. For an interactive agent that's a sharp edge; a large finite ceiling (or a confirmation) would be safer than "run forever".
- **Placement.** The override is applied in the `Event::Key` arm (`main.rs:3146`) rather than where `base_query_config` is consumed, so it re-applies on every keystroke and doesn't run on the `Event::Paste` / `Event::Mouse` arms at all. Applying it at query-build time would be more robust.
- The override is session-only and isn't persisted or surfaced anywhere except a transient status message — worth saying so in the command help text.
- Nits: `PROMPT_SLASH_COMMANDS` is alphabetical and `turns` was appended after `voice`; and "Model & Provider" is an odd help category for a turn limit. No test, and no `docs/commands.md` entry.

One logistics note: I'm landing #403 (the app.rs → `app/` module split) ahead of the feature queue since every open TUI PR needs a revision round anyway. So when you push the fix, please rebase onto main first — `PROMPT_SLASH_COMMANDS`, `help_command_category` and `intercept_slash_command_with_args` will live in `app/commands.rs`, and the new `App` field goes in `app/mod.rs`. Should be a small re-target for a diff this size.

alecuba16:
Reworked per review, rebased onto main (post app/ split — command lives in `app/commands.rs`, field in `app/mod.rs`).

1. **Precedence fixed at the source.** Added `max_turns_override: Option<u32>` to `QueryConfig`, consulted first at the `effective_max_turns` site: override → agent → config. `/turns 50` now works in plan mode.
2. **Applied at query-build time.** `qcfg.max_turns_override = app.max_turns_override` at all four `qcfg` construction sites (key/paste/bridge/remote paths), not in the `Event::Key` arm.
3. **`off` is a finite ceiling (1000),** not `u32::MAX`, so runaway loops still terminate.
4. **Help text** now says session-only, not persisted. Moved to alphabetical position in `PROMPT_SLASH_COMMANDS` and filed under the "Session" category.
5. **`docs/commands.md`** entry added.
6. **Test added:** `max_turns_override_wins_over_agent_definition` — agent pins `max_turns: 20`, config says 2, override 5 wins (6 requests dispatched: 5 turns + summary).

`cargo check --workspace`, `clippy --workspace --all-targets -D warnings`, and the new test pass.

alecuba16:
All review points addressed in `53794e1`:

**Override reaches `effective_max_turns`.** Added `max_turns_override: Option<u32>` to `QueryConfig`. The precedence chain is now `override → agent → config`:
```rust
let effective_max_turns = config.max_turns_override
    .or_else(|| config.agent_definition.as_ref().and_then(|a| a.max_turns))
    .unwrap_or(config.max_turns);
```
So `/turns 50` in plan mode (agent pins `max_turns: Some(20)`) now runs 50 turns, not 20.

**`off` maps to a large finite ceiling (1000), not `u32::MAX`.** A runaway loop still terminates. `MAX_TURNS_CEILING` is defined at the call site.

**Applied at query-build time, not per keystroke.** The override is set on `qcfg.max_turns_override` at all 4 `base_query_config.clone()` sites in `run_interactive` (Enter, Ctrl+Submit, auto-compact, maintenance). Removed from the `Event::Key` arm — no longer re-applies per keystroke, and the `Event::Paste`/`Event::Mouse` arms are covered.

**Session-only, not persisted.** Stated in the command help text (`PROMPT_SLASH_COMMANDS`), the `/turns` handler status messages, and `docs/commands.md`.

**Help category fixed.** `turns` is now in the "Session" category (not "Model & Provider").

**`PROMPT_SLASH_COMMANDS` is alphabetical.** `turns` is placed between `theme` and `ultrareview`.

**`docs/commands.md` entry added.** Documents `/turns`, `/turns N`, `/turns off`, session-local behavior, precedence, the finite ceiling, and degradation interaction.

**Test:** `max_turns_override_wins_over_agent_definition` — agent pins `max_turns: Some(20)`, config says 2, override sets 5; verifies the loop runs past 2 turns and stops at the override cap.

`cargo check --workspace` clean. `cargo clippy --workspace --all-targets` clean. `cargo test -p claurst-query --lib` — 141 passed, 0 failed.

PR #399, opened 2026-08-24
feat(tui): add skills and MCP server count indicator in status bar

alecuba16:
## Summary
Adds skills and MCP server count indicator in the status bar, showing active skills and connected MCP servers.

- Status bar shows skills count and MCP server count
- Clean indicator rendering with dim styling

Clean branch from current main, no shared payload. 24 lines, 2 files.

### DIFF
--- a/src-rust/crates/cli/src/main.rs
+++ b/src-rust/crates/cli/src/main.rs
@@ -1874,6 +1874,13 @@ async fn run_interactive(
     // Set up terminal
     let mut terminal = setup_terminal(live_config.mouse_capture_enabled())?;
     let mut app = App::new(live_config.clone(), cost_tracker.clone());
+    // Status-bar indicator tracks *connected* servers, not the configured
+    // list (a server that fails to handshake is configured but not connected).
+    app.mcp_server_count = tool_ctx
+        .mcp_manager
+        .as_ref()
+        .map(|manager| manager.server_count())
+        .unwrap_or(0);
     if let Some(error) = settings_load_error {
         app.invalid_config_dialog =
             claurst_tui::InvalidConfigDialogState::show_settings_error(&error);
@@ -4194,6 +4201,11 @@ async fn run_interactive(
             let new_mcp_manager = connect_mcp_manager_arc(&decision.allowed).await;
             tool_ctx.mcp_manager = new_mcp_manager.clone();
             app.mcp_manager = new_mcp_manager.clone();
+            // Keep the status-bar indicator in sync with the live count.
+            app.mcp_server_count = new_mcp_manager
+                .as_ref()
+                .map(|manager| manager.server_count())
+                .unwrap_or(0);
             tools_arc = build_tools_with_mcp(new_mcp_manager.clone());
             if app.mcp_view.visible {
                 app.refresh_mcp_view();
--- a/src-rust/crates/tui/src/app/mod.rs
+++ b/src-rust/crates/tui/src/app/mod.rs
@@ -246,6 +246,19 @@ pub struct App {
     pub remote_session_url: Option<String>,
     /// Live MCP manager snapshot source when available.
     pub mcp_manager: Option<Arc<claurst_mcp::McpManager>>,
+    /// Number of connected MCP servers for the status-bar indicator. Updated
+    /// by the caller (CLI) at startup and on every MCP reconnect so it tracks
+    /// `McpManager::server_count()`, not just the configured list.
+    pub mcp_server_count: usize,
+    /// Number of discovered skills for the status-bar indicator. Populated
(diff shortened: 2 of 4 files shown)


Kuberwastaken (changes requested):
Thanks! The wiring itself is correct — the fields land on `App`, `App::new` populates them, and the footer segment renders in the right place between the token-budget and rate-limit blocks.

The blocker is *where* the count is computed. `App::new` is a synchronous constructor called from `crates/cli/src/main.rs:1876` inside the async runtime, and `discover_skills` is not cheap:

- it walks from `cwd` to the filesystem root, scanning `.claurst/skills` and `.agents/skills` at every level (`crates/core/src/skill_discovery.rs:161-172`);
- for every entry in `skills.urls` it calls `fetch_git_skills`, which shells out to `git clone` on a cache miss (`skill_discovery.rs:214-231`).

So for anyone with a `skills.urls` entry, this puts a blocking network fetch on the TUI startup path just to render a number. It also fires in every test that builds an `App` (`crates/tui/src/lib.rs:431`, `crates/tui/src/render.rs:3510` and `:3539`), which both slows the suite and lets the developer's real `~/.claurst/skills` leak into test state.

Two other things:

- **The counts go stale.** They're computed once at startup and never refreshed, so connecting or removing an MCP server leaves the indicator wrong for the rest of the session.
- **`0` is rendered.** The `||` guard means a user with skills but no MCP servers sees `3 skills · 0 mcp`. Worth emitting only the non-zero halves.
- Nit: the new `// Skills + MCP count indicator.` comment sits at column 0 inside an indented block in `render.rs`.

The shape I'd suggest: pass the already-discovered skill count in from the caller rather than re-running discovery inside `App::new`, and update `mcp_server_count` wherever the MCP server list changes. Also heads-up that #401 is adding a `discovered_skills` field on `App` as part of its skill-discovery fix — once that lands you can just read `.len()` off it, which deletes most of this PR.

Sequencing note: I'm landing #403 (the app.rs → `app/` module split) ahead of the feature queue, so please rebase before your next push — the `App` fields and `App::new` body will live in `app/mod.rs`. For a diff this small that should be quick.

alecuba16:
Reworked per review, rebased onto main (post app/ split).

1. **Discovery off the startup path.** `App::new` no longer calls `discover_skills`. The count is computed on a `spawn_blocking` background task and drained through a one-shot channel (`skill_count_pending` / `skill_count_rx`), mirroring the existing `recent_sessions` pattern. `skills.urls` git clones can no longer block TUI startup, and `App::new` in tests stays cheap.
2. **Connected, not configured.** `mcp_server_count` reads `McpManager::server_count()` and is set by the CLI at startup and after every MCP reconnect, so it tracks live connections.
3. **Non-zero halves only.** Footer renders `3 skills`, `2 mcp`, or `3 skills · 2 mcp` — never a `0` half.
4. Comment indent nit fixed.

Happy to rebase onto #401's `discovered_skills` field once it lands to just read `.len()` off it.

`cargo check --workspace` and `clippy --workspace --all-targets -D warnings` clean.

alecuba16:
All review points addressed in `61485c6`:

**Skill discovery moved off `App::new`.** `discover_skills` no longer runs inside the constructor. Instead:
- `skill_count_pending: true` is set on `App::new`, and `skill_count_rx` is `None`.
- On the first run-loop iteration in `app/run.rs`, a one-shot `tokio::spawn` + `spawn_blocking` task runs `discover_skills` in the background and sends the count over an `mpsc` channel.
- The loop drains `try_recv()` each frame and updates `skill_count` when the result arrives.
- No blocking network fetch on the TUI startup path.

**MCP count tracks connected servers, not configured list.** `mcp_server_count` is set from `McpManager::server_count()` at startup in `main.rs`, and updated on every MCP reconnect (the same site that rebuilds `tools_arc`). So connecting or removing an MCP server refreshes the indicator.

**Counts go stale less.** MCP count is refreshed on reconnect. Skill count is a one-shot (discovery is expensive), which matches the review's suggestion. Once #401's `discovered_skills` field lands, the count can be read from `.len()` for live tracking.

**`0` is not rendered.** The indicator only emits non-zero halves: `3 skills` (no `+ 0 mcp`), `2 mcp` (no `0 skills +`), or nothing at all when both are zero. Uses `·` as separator.

**Comment indentation fixed.** The `// 3c. Skills + MCP count indicator` comment is now properly indented inside the block.

`cargo check --workspace` clean. `cargo clippy --workspace --all-targets` clean. `cargo test -p claurst-tui --lib` — 657 passed, 1 pre-existing failure (`settings_screen::tests::all_entries_returns_expected_settings`, unrelated to this PR — fails at HEAD on main too).

Issue #230, opened 2026-07-06
refactor(query): make /goal an in-loop continuation policy + max-steps graceful degradation (MI-3)


Kuberwastaken:
**Priority — VERY IMPORTANT.** Today `/goal` is a cross-turn bolt-on: `check_and_continue_goal` (`crates/query/src/goal_loop.rs:59`) runs after `run_query_loop` returns, re-dispatched from the CLI REPL, so its budget/runaway guards, steering and cancellation straddle two places.

**Fix** — re-express `/goal` as an in-loop continuation policy inside the runner using pi's callback shape (`shouldStopAfterTurn` / `getFollowUpMessages` / `prepareNextTurn`), and adopt opencode's max-steps **graceful degradation**: on hitting `effective_max_turns`, run one final tool-disabled turn that summarizes progress instead of returning cold.

Reference: pi `agent-loop.ts`; opencode `session/runner/max-steps.ts`.

---
_Part of the post-audit hardening pass (deep scan July 2026). Verified against source at the cited lines._

PR #400, opened 2026-08-24
feat(query): add degradationSummaryEnabled config toggle

alecuba16:
## Summary
Adds `degradationSummaryEnabled` config toggle to the query config, allowing users to enable/disable degradation summary behavior.

- Added `degradation_summary_enabled` field to `QueryConfig`
- Wired into config loading and merge logic
- Toggle via `settings.json`

Clean branch from current main, no shared payload. 31 lines, 3 files.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -73,6 +73,7 @@ The `config` object holds runtime behaviour options.
 | `model` | string \| null | provider default | Model ID to use. When absent, the provider's default is used (e.g. `claude-sonnet-4-6` for Anthropic, `gpt-4o` for OpenAI). |
 | `max_tokens` | integer \| null | 8192 | Maximum tokens per model response. |
 | `provider` | string \| null | `"anthropic"` | Active provider. See the [Providers](#providers) section. |
+| `degradationSummaryEnabled` | boolean \| null | true | Controls the max-steps graceful degradation turn. When `true` (or unset), exceeding `max_turns` runs one final tool-less turn asking the model to summarize progress. When `false`, the loop returns the last assistant message immediately. |
 
 ### Permission mode
 
--- a/src-rust/crates/core/src/lib.rs
+++ b/src-rust/crates/core/src/lib.rs
@@ -1114,6 +1114,12 @@ pub mod config {
         /// Keyboard scrolling (PageUp/PageDown, etc.) is unaffected either way.
         #[serde(default, rename = "mouseCapture", skip_serializing_if = "Option::is_none")]
         pub mouse_capture: Option<bool>,
+        /// Whether the max-steps graceful degradation summary turn is enabled.
+        /// `None` (default) or `Some(true)` means exceeding `max_turns` runs one
+        /// final tool-less turn asking the model to summarize progress.
+        /// `Some(false)` returns cold (last assistant message) immediately.
+        #[serde(default, rename = "degradationSummaryEnabled", skip_serializing_if = "Option::is_none")]
+        pub degradation_summary_enabled: Option<bool>,
     }
 
     #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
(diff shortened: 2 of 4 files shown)


Kuberwastaken (changes requested):
Thanks for this — an escape hatch for the max-steps degradation turn is reasonable, and the query-loop side is wired correctly (the cold-return branch mirrors the existing `degradation_done` path exactly, which I appreciate).

Two things block it as written, both in the config plumbing:

**1. The default is actually `false`, not `true`.** `Config` derives `Default` (`crates/core/src/lib.rs:1015`), and `#[serde(default = "default_true")]` only applies when *deserializing*. `Settings::load_from_path` returns `Self::default()` when the file doesn't exist (`crates/core/src/lib.rs:1728`), so a fresh install with no `settings.json` gets `degradation_summary_enabled: false` — silently turning off the #230 degradation summary for everyone who hasn't written a settings file. That contradicts the doc comment on the field.

**2. The merge operator inverts the intent.** In `Settings::merge` you added:
```rust
degradation_summary_enabled: over.config.degradation_summary_enabled && base.config.degradation_summary_enabled,
```
Every other bool in that block uses `||` (`file_injection_enabled`, `cursor_blink_enabled`, `disable_claude_mds`, …). With `&&` plus the derived-`Default` `false` above, a project `.claurst/settings.json` alongside a missing global settings file resolves to `false && true` → `false`, so the flag latches off. It also means a project-level `true` can't re-enable a global `false`.

Both problems disappear if you model it the way the neighbouring tri-state flags are modelled — `mouse_capture` and `auto_commits`, two lines up:

```rust
#[serde(default, rename = "degradationSummaryEnabled", skip_serializing_if = "Option::is_none")]
pub degradation_summary_enabled: Option<bool>,
// merge:
degradation_summary_enabled: over.config.degradation_summary_enabled.or(base.config.degradation_summary_enabled),
// read site:
degradation_enabled: cfg.degradation_summary_enabled.unwrap_or(true),
```

`None` then genuinely means "unset → default true", and project-over-global precedence matches every other option.

Two smaller asks:
- A test covering the disabled path (turn cap exceeded → `EndTurn` without the summary turn) — the new branch is currently untested.
- A line in `docs/configuration.md`; new settings keys are documented there.

Happy to merge once the flag is `Option<bool>` with `.or()` merging — this one's isolated to `core`/`query`, so no rebase pain from the other queued work.

alecuba16:
Both blockers fixed, rebased onto main post app/ split.

1. `degradation_summary_enabled` is now `Option<bool>` with `#[serde(default, skip_serializing_if = "Option::is_none")]` — `None` genuinely means unset, so fresh installs keep the degradation summary.
2. Merge is `over.config.degradation_summary_enabled.or(base.config.degradation_summary_enabled)`, matching `mouse_capture`/`auto_commits`. A project-level `true` can now re-enable a global `false`.
3. Read site: `QueryConfig::from_config*` uses `.unwrap_or(true)`.
4. Added `max_steps_disabled_returns_cold_without_summary_turn` — turn cap exceeded with the flag off dispatches no extra request and injects no summary prompt.
5. Added `test_config_degradation_summary_merge_precedence` — tempdir + `CLAURST_HOME` isolation (no real `~/.claurst/settings.json` read), covering project-wins, global-applies, and unset-defaults-on.
6. `docs/configuration.md` documents the key.

`cargo check --workspace` and `clippy --workspace --all-targets -D warnings` clean.

alecuba16:
Both blockers fixed in `9ee9173`:

**1. Flag is now `Option<bool>` with `.or()` merging.** Changed from `bool` to `Option<bool>` in `Config`, matching `mouse_capture` and `auto_commits`. Merge uses `over.config.degradation_summary_enabled.or(base.config.degradation_summary_enabled)` — project value wins, unset falls back to global, global unset falls back to `None`. Read site uses `cfg.degradation_summary_enabled.unwrap_or(true)`, so a fresh install with no settings.json keeps the degradation summary on.

**2. Merge no longer inverts intent.** The `&&` operator is replaced with `.or()`, consistent with every other tri-state option in that block.

**Tests:**
- `test_config_degradation_summary_defaults_on`: verifies `Config::default()` gives `None` and `.unwrap_or(true)` yields `true`.
- `test_config_degradation_summary_merge_precedence`: project `true` + global `false` → `true` (project wins); project unset + global `true` → `true` (fallback); both unset → `None` → `true` (default). Uses `tempfile::tempdir()` + `CLAURST_HOME` isolation, no real `~/.claurst` reads.
- `max_steps_disabled_returns_cold_without_summary_turn`: drives the query loop with `degradation_enabled: false`, verifies it returns `EndTurn` (cold, last assistant message) without running the summary turn.

**Docs:** `docs/configuration.md` updated with the new `degradationSummaryEnabled` key.

`cargo check --workspace` clean. `cargo clippy --workspace --all-targets` clean. `cargo test -p claurst-core -p claurst-query --lib` — 645 passed, 0 failed.

PR #403, opened 2026-09-02
refactor(tui): split 7.8k-line app.rs into cohesive app/ modules

SlugThug:
From SlugThugLabs — building from whatever's on hand, to standards worth aspiring to.

## Summary

`tui/src/app.rs` was a 7,807-line file containing the full `App` state struct, a 5,557-line impl block, the keyboard dispatch tree, and a 950-line test suite. This PR restructures it into `app/` with one module per responsibility. **Pure structural move — zero behavior change.** Every function body is byte-identical to the original; no logic was touched, no warnings fixed, no style "improved" while moving.

## Module layout

| Module | Responsibility | Lines |
|---|---|---|
| `mod.rs` | `App` struct, constructor, re-exports, module wiring | ~940 |
| `types.rs` | Pure data model (messages, tool blocks, dialogs, state) | 245 |
| `keys.rs` | Keyboard normalization + key-event dispatch | 2,727 |
| `mouse.rs` | Pointer input, selection, context menu, paste viewer | 660 |
| `providers.rs` | Provider/model/config + import-config flow | 442 |
| `commands.rs` | Slash-command catalog + intercept dispatch | 392 |
| `turns.rs` | Turn lifecycle (snapshots, streaming flush, labels) | 176 |
| `messages.rs` | Transcript ops (add/replace/push, notifications, scroll) | 162 |
| `prompt.rs` | Prompt input state machine + voice hold-to-talk | 158 |
| `views.rs` | Overlay/dialog orchestration + MCP view state | 405 |
| `run.rs` | Main event loop, query events, jump-to-error | 628 |
| `tests.rs` | Test suite (moved whole, `use super::*`) | 956 |

Public API unchanged — `crate::app::{App, ...}` re-exported from `mod.rs` exactly as before; zero ripples into `render.rs`, `messages/`, or `lib.rs`.

## Visibility note (the one deliberate adjustment)

Private methods moved across module boundaries became `pub(super)` — restoring exactly the visibility the single-module layout provided. Of the 79 items lifted: **48 have genuine cross-module callers; 31 are used only within their defining file** and are candidates to return to `private` during the phase-two interface work below.

## Placement decisions

- `try_copy_to_clipboard` stays in `mod.rs` — it's crate-level public API, re-exported by `lib.rs` (`pub use app::try_copy_to_clipboard`).
- Animation/mode helpers (`tick_rustle_pose`, `cycle_agent_mode`, `apply_theme`, …) stay near the constructor for now; their callers live inside `app`, and regrouping them belongs with the interface work.
- MCP plumbing in `views.rs` mirrors the original file's conflation; splitting it out is flagged for phase two.

## Known follow-up (phase two, separate PRs)

Honest caveats this PR does **not** hide:

- `keys.rs` remains the largest file — `handle_key_event` alone is ~2,300 lines and is the next target.
- Submodule impl blocks still share the full `App` state; phase two replaces `&mut self`-everywhere with narrow interfaces and encapsulated state, and privatizes the ~31 lazy lifts.
- `cargo fmt` stays advisory per repo policy — import ordering here matches the crate's existing conventions (`messages/`, etc.); fmt drift is at parity with upstream (156 diffs pre-split → 171 across the same volume).

## Verification (matches CI gates)

- `cargo check --workspace --locked` ✅
- `cargo test --workspace --locked` ✅ 1,799 tests, 0 failures (tui: 708)
- `cargo clippy --workspace --all-targets -D warnings` ✅ clean
- 208/208 functions preserved vs. original — 0 missing, 0 extra
- Diff: `app.rs` deleted, 11 new files, **+7,893 / −7,807** (the +86 is module glue)

No behavior change, no test changes, no dependency changes.


Kuberwastaken:
Reviewed this closely and it holds up — I diffed all 208 functions against `app.rs` and every body is byte-identical, with the only signature changes being the 79 `pub(super)` lifts you documented. Item and attribute inventories match exactly (23 structs/enums/consts, all 10 `#[cfg]` blocks), `#[cfg(test)]` gating is preserved on `mod tests`, and the import set is 1:1 after the flattening. CI is green on all three platforms. Really nice work, and thanks for the honest caveats in the description.

One small correction to "Public API unchanged": `ContextMenuState`, `ContextMenuItem`, and `GoToLineDialog` are `pub` in today's `app.rs` and reachable as `claurst_tui::app::*`, but they're missing from the `pub use types::{…}` re-export in `mod.rs`. Nothing in the workspace consumes them so it builds fine — I'll restore them in a follow-up commit rather than send this back for a one-liner.

On timing: seven open PRs currently touch `app.rs`, but every one of them needs a revision round anyway, so rather than make you rebase behind a queue that isn't moving, I'm landing this first — the module layout becomes the base everyone else re-targets. Merging now. 🎉

